PR #23, opened 2014-11-19
Set GPU example's make file to explicitly use python3 and fix aemstro enviroment variable

WhiteDawn:
Before, the make file would use the system default python, aemstro only work with python3 so explicitly calling python3 in the make file will help avoid confusion.

Changed aemstro's path enviroment variable from AEMSTRO to AEMSTROPATH because the readme file defines it as AEMSTROPATH and I believe AEMSTROPATH is more descriptive

### DIFF
--- a/examples/gpu/Makefile
+++ b/examples/gpu/Makefile
@@ -161,7 +161,7 @@ $(OUTPUT).elf	:	$(OFILES)
 %.vsh.o	:	%.vsh
 #---------------------------------------------------------------------------------
 	@echo $(notdir $<)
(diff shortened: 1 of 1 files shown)


WinterMute:
The Makefile has been updated to use python3 explicitly but I don't see any reason why the aemstro variable should be renamed. In any case we're going to be moving away from aemstro for compiling shaders

profi200:
That would be great if it no longer depends on Python. But please don't go with one of these native shader assemblers which are blown with something like boost. That's a nightmare for users which want/need to compile it.

PR #54, opened 2014-12-13
Fixed corrupted screen using single buffer

Aurelio92:
Made a little change in gfxSetDoubleBuffering() to avoid graphics being
corrupted when using single buffer mode and never calling
gfxSwapBuffers()


smealum:
dismissed ! ﾟヽ(｡☉౪ ⊙♥)ﾉﾟ

PR #62, opened 2014-12-14
Added Pad example

Aurelio92:
This example shows how to get keys info


WinterMute:
merged into circle pad example

PR #65, opened 2014-12-14
Added RTC example

Aurelio92:
This example shows how to get os time and prints it on top screen


linoma:
is it ready? :dancers:

Aurelio92:
No. Meanwhile you can use the osGetTime function just as it is shown in this example

WinterMute:
time now works. Sorry to take so long - it was dependent on some things in devkitARM r44.

Could you update for latest libctru - accounting for already initialised services etc. Also use $(MAKE) in the makefiles rather than using make directly. That allows things to work on BSD where gnu make is known as gmake & also passes on -j parameters.

linoma:
ooohhh yeah ;)

smealum:
so does that mean we can close this now ?

archshift:
Note that this won't even build with the latest libctru.

Aurelio92:
Damn I still have to update this example... I'll do it right now

smealum:
lgtm

PR #79, opened 2015-01-08
Added missing declarations in csnd header file

master4523:
Some functions declarations were missing in csnd.h


fincs:
libctru's CSND code is right now being [refactored](https://github.com/smealum/ctrulib/tree/better-csnd).

PR #83, opened 2015-01-16
Get rid of spaces.

profi200:


fincs:
Tabs IMO shouldn't be used for alignment, that's what spaces are for.

profi200:
Meh, the entire shit again.

PR #88, opened 2015-01-19
Add AM_GetTitleProductCode, FSUSER_GetNandArchiveResource, and fix missing service imports in 3ds.h.

Steveice10:
Adds AM_GetTitleProductCode to the am services and FSUSER_GetNandArchiveResource to the fs services.

Also adds missing service header includes to 3ds.h.

### DIFF
--- a/libctru/include/3ds/services/am.h
+++ b/libctru/include/3ds/services/am.h
@@ -91,3 +91,13 @@ Note: The title must have the uniqueid: 0x00000, otherwise this will fail.
   titleid		title id of title
 */
 Result AM_InstallFIRM(u8 mediatype, u64 titleid);
+
+/* AM_GetTitleProductCode()
+About: Gets the product code of a title based on its title id.
+
+  mediatype		mediatype of title
+  titleid		title id of title
+  productcode           buffer to output the product code to (should have a length of 16) 
+*/
+Result AM_GetTitleProductCode(u8 mediatype, u64 titleid, char* productcode);
+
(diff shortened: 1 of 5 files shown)


Rinnegatamante:
Product code is not only 4 bytes related but it's 16 bytes ( http://3dbrew.org/wiki/AM:GetTitleProductCode )... If you taken this code from my Big Red Menu ( https://github.com/Rinnegatamante/Big_Red_Menu/blob/master/source/luaSystem.cpp ), this is why i use a sprintf and not a snprintf...

Steveice10:
Ahh, you're right. I read "2-5" as bytes 2-5, forgetting that each index was 4 bytes. Will fix.

Steveice10:
Fixed a missing include and added missing service headers to 3ds.h.

spinda:
Any chance this could be merged in?

libctru/include/3ds/services/am.h
```diff
@@ -90,4 +90,13 @@ Note: The title must have the uniqueid: 0x00000, otherwise this will fail.
...
+Result AM_InstallFIRM(u8 mediatype, u64 titleid);
+
+/* AM_GetTitleProductCode()
+About: Gets the product code of a title based on its title id.
+
+  mediatype		mediatype of title
+  titleid		title id of title
+  productcode           buffer to output the product code to (should have a length of 16) 
+*/
+Result AM_GetTitleProductCode(u8 mediatype, u64 titleid, char* productcode);
```
archshift: newline at the end

Steveice10: Fixed.

profi200:
See this: https://github.com/smealum/ctrulib/pull/99

Waiting for it to be merged.

PR #93, opened 2015-02-02
HTTPC_AddRequestHeaderField Fix

LunarCookies:


WinterMute:
why 12?

LunarCookies:
It's L2, the strlen + 1 of the value parameter as documented here: http://3dbrew.org/wiki/HTTPC:AddRequestHeader

WinterMute:
why is the length of the value parameter 12?

LunarCookies:
It's not a number, it's a variable name: L2 (in lowercase)

WinterMute:
can you change those variables to some sensible name while you're there, please?

mtheall:
I vote name_len and value_len.

LunarCookies:
Those was the names provided by the original developer I suppose it's not by to me to change them

mtheall:
We just mean change the names while you're changing the related code anyway.

LunarCookies:
Yes I know, but the rest of the functions use the same nomenclature when naming variables. That's why I  am reticent about changing them.

WinterMute:
Awesome, cheers.

PR #95, opened 2015-02-22
updated shader for current aemstro

elhobbs:
updated shader for the gpu example that  works with the current aemstro version


smealum:
great, have you tested it ?

elhobbs:
yes, I tested it. I did not notice the lighting was not working correctly. It is fixed now.

smealum:
ok cool, thanks.

PR #96, opened 2015-02-25
Add svcLockMutex

Lectem:
Just a shortcut, since most people probably want to lock a mutex for good until it's released


mtheall:
That timeout is now how long to keep it locked. It's how long to wait until you can acquire it. If it times out then you didn't acquire the lock.

mtheall:
s/now/not/

Lectem:
According to 3dbrew, the time is in nanoseconds, so it would make it about.. 584years.

But I think that the best reason is the following :
If we look at https://msdn.microsoft.com/library/cc190477%28v=vs.110%29.aspx and suppose it works the same way, then -1 should lock indefinitely the mutex. U64_MAX is exactly that (0xffffffffffffffffULL == -1 using two's complement representation)

plutooo:
Well, this is functionally correct but kinda confusing.

PR #98, opened 2015-03-01
Improve shader support

neobrain:
* output attribute component masks are considered now
* boolean and integer uniforms can be set now

I didn't rename existing functions to be consistent because that would break backwards compatibilty. I'll leave it up to someone else to rename them, hence.

### DIFF
--- a/libctru/include/3ds/gpu/gpu.h
+++ b/libctru/include/3ds/gpu/gpu.h
@@ -190,6 +190,8 @@ typedef enum{
 }GPU_Primitive_t;
 
 void GPU_SetUniform(u32 startreg, u32* data, u32 numreg);
+void GPU_SetIntUniform(u32 startreg, u32 value);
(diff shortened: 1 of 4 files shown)


WinterMute:
This is redundant, see https://github.com/smealum/ctrulib/tree/gpu_revamp

fincs:
Further GPU work should go into the gpu_revamp branch as WinterMute pointed out. Additionally we do not worry about breaking backwards compatibility.

neobrain:
Heh, thanks for the pointers.

neobrain:
@WinterMute Can you explain what you meant with "redundant"? As far as I can see the functionality added in this PR is not present either in gpu_revamp. Should I port the code over to that branch?

fincs:
See commit 05e886c9fa8bf37a08586974baea1de44b97be13.

neobrain:
I see. Thanks again for clearing that up.

PR #99, opened 2015-03-01
Added missing am service cmds and changed the names to reflect 3dbrews d...

profi200:
...ocs.

All added functions are tested.

### DIFF
--- a/libctru/include/3ds.h
+++ b/libctru/include/3ds.h
@@ -13,8 +13,10 @@ extern "C" {
 #include <3ds/os.h>
 #include <3ds/gfx.h>
 #include <3ds/console.h>
+#include <3ds/util/utf.h>
 
 #include <3ds/services/ac.h>
+#include <3ds/services/am.h>
 #include <3ds/services/apt.h>
 #include <3ds/services/cfgnor.h>
 #include <3ds/services/cfgu.h>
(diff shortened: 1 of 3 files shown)


WinterMute:
Please don't mix whitespace changes with code changes.

I'm not convinced by the function name changes either but diffs get really hard to read if you both rename a function and move it's position so far.

profi200:
https://github.com/profi200/ctrulib/blob/master/libctru/include/3ds/services/am.h#L51-58
https://github.com/profi200/ctrulib/blob/master/libctru/include/3ds/services/am.h#L61-69

https://github.com/profi200/ctrulib/blob/master/libctru/source/services/am.c#L68-83
https://github.com/profi200/ctrulib/blob/master/libctru/source/services/am.c#L85-101

The other things are just position/name changes. I think it's better to use the 3dbrew names.

WinterMute:
3dbrew can change to reflect libcrtu naming, there's no particular reason for slavishly sticking to what it currently says on 3dbrew. Especially if that means a function gets called AM_StartInstallCIADB1 rather than AM_StartDlpChildCiaInstall

profi200:
The description in the header file clearly says what it does.
And that's actually what it does. It installs the title in DB 1.

yellows8:
"It installs the title in DB 1" Which is the same as "DlpChild", aka the tmp DB.

PR #100, opened 2015-03-01
Added AM_ListTitles(), removed nonsense bitwise ands, replaced snprint() ...

profi200:
...with memcpy() (overflow safe) and made utf.h aveilable for apps.

### DIFF
--- a/libctru/source/services/am.c
+++ b/libctru/source/services/am.c
@@ -199,7 +199,7 @@ Result AM_GetTitleProductCode(u8 mediatype, u64 titleID, char* productCode)
 	if((ret = svcSendSyncRequest(amHandle))!=0) return ret;
 
 	// The product code string can use the full 16 bytes without NULL terminator
-	if(productCode) memcpy(productCode, &cmdbuf[2], 16);
+	if(productCode) snprintf(productCode, 16, "%s", (char*)&cmdbuf[2]);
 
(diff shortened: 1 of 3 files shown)


profi200:
Have not read it carefully. snprintf() does the same as this. It's just slower probably.

libctru/source/services/am.c
```diff
@@ -89,96 +108,97 @@ Result AM_StartDlpChildCiaInstall(Handle *ciahandle)
...
-	cmdbuf[2] = titleid & 0xffffffff;
-	cmdbuf[3] = (titleid >> 32) & 0xffffffff;
+	cmdbuf[2] = titleID & 0xffffffff;
+	cmdbuf[3] = (u32)(titleID >> 32);
 	
 	if((ret = svcSendSyncRequest(amHandle))!=0) return ret;
 
-	snprintf(productcode, 16, "%s", (char*)(&cmdbuf[2]));
+	// The product code string can use the full 16 bytes without NULL terminator
+	if(productCode) snprintf(productCode, 16, "%s", (char*)&cmdbuf[2]);
```
WinterMute: Being honest I think I prefer the memcpy. Is there an advantage to using snprintf here?

profi200: Other than an automatically added NULL terminator, nothing. snprintf() is also slower i think.

PR #103, opened 2015-03-09
add stubs for svc 0x61, 0x63, 0x64

patois:


fincs:
This patch does not describe the full DebugEventInfo structure (which is described in detail in the Wiki). Moreover, there's a stray trailing whitespace change in getThreadCommandBuffer().

PR #107, opened 2015-03-23
Hello triangle example

Lectem:
I tried to reduce the GPU example to the minimum, and give a bit more details.
Every comment might not be accurate, please correct me if I'm wrong somewhere.
It uses Picasso for the shader (the makefile assumes it is in your PATH).


Lectem:
I squashed the commits and updated the code to make a better use of our current knowledge about GX and the Pica, and use the latest enums provided by https://github.com/smealum/ctrulib/pull/112

fincs:
See commit bf7d686e88.

PR #108, opened 2015-03-24
Added linearCalloc.

Rinnegatamante:
Added calloc implementation for linear heap and vram.


profi200:
For linear mem you can use memset().
For VRAM this will not work because apps usually have no direct write access to VRAM.

smealum:
agree with profi. some apps do have direct vram access but not all, so i don't thin vramcalloc in its current form is a good idea. you could do it with a mem fill GX call... but then the user would have to be aware that it's asynchronous, and i think maybe that should be made explicit. vramCallocAsync maybe ? idk doesn't have that nice a ring to it. maybe just cut vramCalloc altogether.

fincs:
I think neither linearCalloc nor vramCalloc are worth it. Users can always zerofill the memory themselves.

PR #109, opened 2015-03-25
Reset GPUCMD buffer after running GPU_Reset commands.

Steveice10:
GPU_Reset makes various GPU buffer writes, however it doesn't set the buffer offset back to 0 after flushing and running. Because of this, the GPU example would freeze on a black screen for some people. Setting the buffer offset to 0 at the end of GPU_Reset fixes this issue.

Why this issue only occurred for some people, I don't quite know. However, I have tested this fix and can confirm that it works.


filfat:
Works for me (N3DS 9.0E)

PR #110, opened 2015-03-28
Define SVC functions as inlined functions

173210:
This commit reduces overheads to call SVC functions.
This trick is used also on Linux kernel. I hope my commit will be utilized for implementing other SVCs.

### DIFF
--- a/libctru/include/3ds/svc.h
+++ b/libctru/include/3ds/svc.h
@@ -188,51 +188,366 @@ static inline u32* getThreadCommandBuffer(void)
 	return (u32*)((u8*)getThreadLocalStorage() + 0x80);
 }
 
+static inline void __attribute__((noreturn)) svcExitProcess()
+{
+	asm volatile("svc 0x03"
+		::: "lr");
+	__builtin_unreachable();
+}
+
+static inline void __attribute__((noreturn)) svcExitThread()
+{
+	asm volatile("svc 0x09"
+		::: "lr");
+	__builtin_unreachable();
+}
+
+static inline void svcSleepThread(s64 ns)
+{
+	register s64 r0 asm("r0") = ns;
+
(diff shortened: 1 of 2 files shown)


fincs:
No, this is extremely brittle and it is guaranteed to produce problems with the compiler.

profi200:
It's also not good pratice to place code in headers.

fincs:
Err, placing code in headers is necessary in order for inlined functions to work.

profi200:
Not really. You can still just place the prototype in the header and the code in another file. The compiler will accept both ways.

173210:
It highly depends on GCC. It may not work on another compilers. But it won't be a problem because it had already had asm and attributes statements which depends on GCC.
To see how it depends on GCC, compile with -pedantic option.
To see output code, disassemble with objdump.

173210:
fincs, you may worry about optimization, but you don't have to.
asm statements which I wrote includes information to optimize code properly.
In the detail, all asm statement has information about registers which are used to input arguments/ouput results/call SVC.

WinterMute:
@173210 asm is a gnu extension which will cause errors with -ansi and various -std options. It would be better to use __asm__ for these. Is adding lr to the clobber list really necessary? I was under the impression that these calls couldn't be used from supervisor mode.

Did you do any checks to see what registers other than input & output are changed, if any?

@profi200 it's not good practice to place non inlined code in headers, static inline code is perfectly acceptable. I'm also reasonably sure that you have to enable link time optimisation to get code in separate translation units to inline.

173210:
I think we should use __asm__ for header files, but I used asm because asm is already used. I think we should fix them with another commit.
Well, actually LR is not necessary. I'll fix them.

173210:
I'll close this commit because WinterMute found a problem.

PR #112, opened 2015-04-03
GX_SetDisplayTransfer enums, GPU_SetTexture bug, GPU sample commented

xerpi:
Added GX_SetDisplayTransfer flags enums and define helpers
Solved GPU_SetTexture bug (width and height was swapped)
Comment the GPU sample a bit more (to make it easier to understand)

### DIFF
--- a/libctru/include/3ds/gpu/gx.h
+++ b/libctru/include/3ds/gpu/gx.h
@@ -2,6 +2,29 @@
 
 #define GX_BUFFER_DIM(w, h) (((h)<<16)|((w)&0xFFFF))
 
+typedef enum
+{
+	GX_TRANSFER_FMT_RGBA8  = 0,
+	GX_TRANSFER_FMT_RGB8   = 1,
+	GX_TRANSFER_FMT_RGB565 = 2,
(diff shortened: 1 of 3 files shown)


libctru/include/3ds/gpu/gx.h
```diff
@@ -2,6 +2,29 @@
...
+	GX_TRANSFER_FMT_RGB565 = 2,
+	GX_TRANSFER_FMT_RGB5A1 = 3,
+	GX_TRANSFER_FMT_RGBA4  = 4
+} GX_TRANSFER_FORMAT;
+
+typedef enum
+{
+	GX_TRANSFER_SCALE_NO = 0,
+	GX_TRANSFER_SCALE_X  = 1,
+	GX_TRANSFER_SCALE_Y  = 2
```
aliaspider: doesn't this flag scale down both X and Y ?
also the flags after this line lack the truncate bit, but that can be added later I guess.

xerpi: Yeah you are right, some bits are missing.

PR #113, opened 2015-04-03
Implement svcBackdoor in SVCs

archshift:


profi200:
Why not name it svcEnterKernelmode()? How should newbies know what this does otherwise?

fincs:
svcBackdoor is the "official" name for this syscall. Also, I'm not sure if it's a good idea to add it to libctru, especially since a previous exploit is necessary for its use.

profi200:
Backdoor is still misleading. And it's not the official name just because someone named it like that on 3dbrew.

archshift:
@fincs, the same can be said for many of the SVCs in libctru, which cannot be used normally with any of the usermode exploits we have.

@profi200, backdoor is IMO a pretty good name for it: it's a backdoor into the ARM11 kernel. I wouldn't think it's misleading, and plenty of the SVC names are indecipherable to beginners already (svcArbitrateAddress comes to mind).

smealum:
there's no point in changing the name unless it's also changed on 3dbrew, and there's no points in changing it on 3dbrew unless the official name somehow comes out. backdoor is descriptive enough.
and yeah at this point no real reason to not just have all svc calls in ctrulib.

PR #115, opened 2015-04-14
Add news:u commands.

Steveice10:
Adds the commands from news:u, which is only a single command, NEWSU_AddNotification. This command has been tested with the following input:

    NEWSU_AddNotification(u"Hello world!", strlen("Hello world!"), u"The world would like to say hello back!", strlen("The world would like to say hello back!"), NULL, 0, false);

Producing the following result:

Result


smealum:
lgtm

PR #117, opened 2015-04-15
svc.h: Improve compatiblity

173210:
Refer WinterMute's comment in this pull request.
https://github.com/smealum/ctrulib/pull/110

### DIFF
--- a/libctru/include/3ds/svc.h
+++ b/libctru/include/3ds/svc.h
@@ -178,7 +178,7 @@ typedef struct {
 static inline void* getThreadLocalStorage(void)
 {
 	void* ret;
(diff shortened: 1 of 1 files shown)


WinterMute:
you don't need to use \__volatile__, the volatile keyword exists with -ansi, -pedantic & all the -std options I've tried.

173210:
I used __volatile__ because other header files use it.

yuriks:
I don't think this particular asm snippet even need the volatile keyword: it has no side effects apart from the output register. This means that omitting the volatile is safe.

173210:
That's right, I'll open another pull request.

173210:
Please merge #123.

PR #120, opened 2015-05-16
added missing texenv op

Lectem:
Texture environment operands enums were missing, added the values based on 3dbrew (http://3dbrew.org/wiki/GPU_Commands#Parameter_structure_for_command_0x00C0_2)


smealum:
GL prefixes ? why ?

Lectem:
My bad, I was too deep in the OGL specification and forgot ctrulib used the GPU prefix, gonna fix it

PR #122, opened 2015-05-25
threads svcs

Lectem:
I added most of the svcs related to threads

### DIFF
--- a/libctru/source/svc.s
+++ b/libctru/source/svc.s
@@ -59,10 +59,46 @@ svcSleepThread:
 	svc 0x0A
 	bx  lr
 
+.global svcGetThreadPriority
+.type svcGetThreadPriority, %function
+svcGetThreadPriority:
+	str r0, [sp, #-0x4]!
+	svc 0x0B
+	ldr r3, [sp], #4
+	str r1, [r3]
+	bx  lr
+	
 .global svcSetThreadPriority
 .type svcSetThreadPriority, %function
 svcSetThreadPriority:
 	svc 0x0C
+	bx  lr
+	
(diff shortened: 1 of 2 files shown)


smealum:
LGTM

libctru/source/svc.s
```diff
@@ -259,6 +295,17 @@ svcGetProcessInfo:
 	ldr  r4, [sp], #4
 	bx   lr
 
+.global svcGetThreadInfo
+.type svcGetThreadInfo, %function
+svcGetThreadInfo:
```
smealum: why not use r3

Lectem: As I said on irc, I merely followed what was done for svcGetProcessInfo. I didn't test this one since we don't know what values we should get depending the the 3rd argument.
If it bothers you I'll comment it out

smealum: no need to comment it out... just don't push/pop r4 and use r3 instead.

Lectem:
Ok so I took the liberty to change all the svcGet*Info and shorten the implementations.
However, note that svcGetThreadInfo seems to always return error 0xd8e007ed (untested for the other svcGet*Info).

fincs:
:+1:

plutooo:
Are you sure that your modification to svcGetSystemInfo is correct? Because previous implementation wrote 3x u32's, whereas your new one only writes 2x u32's.

Lectem:
I just tested svcGetSystemInfo, it works fine.

plutooo:
That wasn't my question, but yes I looked it up and apparently the old implementation was wrong.

PR #123, opened 2015-05-26
svc.h: Remove volatile for getThreadLocalStorage

173210:
Refer to comments in pull request #117.

### DIFF
--- a/libctru/include/3ds/svc.h
+++ b/libctru/include/3ds/svc.h
@@ -178,7 +178,7 @@ typedef struct {
 static inline void* getThreadLocalStorage(void)
 {
 	void* ret;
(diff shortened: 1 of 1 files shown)


smealum:
didn't @WinterMute's comment on #110 say we should use __asm__ ?

173210:
I'll fix with another commit. Please merge this at first.

smealum:
i think you can just add another commit to this same PR

173210:
That's right. But I don't have time to add commits now. I'll add other commits when I'm free.

smealum:
i'll just merge and make an inline commit, thanks

173210:
Thank you.

smealum:
done, should be fine now.

PR #130, opened 2015-06-27
svc.h cleanup : enums in structs and fix indent and reorder

Lectem:
Using the enums in structs instead of u32 (and forcing size using bitfields).
Also reordered stuff to be more coherent for a future documentation.
Added the types.h include in case someone is willing to include it without including 3ds.h

### DIFF
--- a/libctru/include/3ds/svc.h
+++ b/libctru/include/3ds/svc.h
@@ -4,6 +4,8 @@
 
 #pragma once
 
+#include "types.h"
+
 typedef enum {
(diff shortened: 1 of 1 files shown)


fincs:
I think using the enum types instead of u32 in the structs is unnecessary and it leads to requiring ugly hacks to make sure the enum type is treated as a 32-bit type.

Lectem:
It's been discussed on IRC and we also came to this conclusion.

PR #135, opened 2015-07-07
updated sound example for new API

Lectem:
The sound example was still using the old CSND API, so I updated it.
I also added some text output so that it is easier to know what's going on when running the example.

### DIFF
--- a/examples/audio/mic/source/main.c
+++ b/examples/audio/mic/source/main.c
@@ -14,8 +14,13 @@ int main()
 	u32 audio_initialized = 0;
 
 	gfxInitDefault();
+	consoleInit(GFX_BOTTOM, NULL);
 
-	if(CSND_initialize(NULL)==0)audio_initialized = 1;
+	if(csndInit()==0)
+	{
+		printf("Init success\n");
+		audio_initialized = 1;
+	}
 
 	sharedmem = (u32*)memalign(0x1000, sharedmem_size);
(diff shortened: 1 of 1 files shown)


examples/audio/mic/source/main.c
```diff
@@ -57,9 +63,10 @@ int main()
 
 			if(hidKeysUp() & KEY_A)
 			{
+				printf("Playing the recorded sample\n");
 				MIC_SetRecording(0);
 				GSPGPU_FlushDataCache(NULL, audiobuf, audiobuf_pos);
-				CSND_playsound(0x8, CSND_LOOP_DISABLE, CSND_ENCODING_PCM16, 16000, (u32*)audiobuf, NULL, audiobuf_pos, 2, 0);
+				csndPlaySound(0x8, SOUND_LOOPMODE(CSND_LOOPMODE_MANUAL)|SOUND_FORMAT(CSND_ENCODING_PCM16), 16000, 2,0,(u32*)audiobuf, NULL, audiobuf_pos);
```
fincs: Should have been:
```c
csndPlaySound(0x8, SOUND_ONE_SHOT | SOUND_FORMAT_16BIT, 16000, 1.0, 0.0, (u32*)audiobuf, NULL, audiobuf_pos);
```
Other than that, it's good.

Lectem: Fixed.

PR #136, opened 2015-07-07
added texture border color

Lectem:
I am not sure if it should be part of GPU_SetTexture or if we should have a separate call for it, waiting for feedback.


smealum:
i want to say it should be a separate function. it's not something most people are going to want to update each time they bind a texture so as it is all it would do is waste processing time. (admittedly not much, but still...)

Lectem:
I was thinking most people would indeed not use it, and it takes some command buffer memory, so I suppose I'll just make it a separate call.

fincs:
:+1: to having a separate function.

PR #143, opened 2015-07-24
Fix GPU_Reset having wrong GSP command buffer

aroulin:
```void GPUCMD_Run(u32* gxbuf)``` wants gxbuf, i.e. the GSP command buffer, as argument but is given the GPU command buffer ```gpuBuf``` instead.

I have tracked down the use of this ```gxbuf``` argument until ```GX_SetCommandList_First``` and ```GSPGPU_SubmitGxCommand``` and it does not seem it is a misnamed variable.

I am not able for now to test this fix on real hardware.

### DIFF
--- a/libctru/source/gpu/gpu.c
+++ b/libctru/source/gpu/gpu.c
@@ -215,7 +215,7 @@ void GPU_Reset(u32* gxbuf, u32* gpuBuf, u32 gpuBufSize)
 	for(i=0;i<gpuResetSequenceLength;i++)GPUCMD_AddSingleParam(gpuResetSequence[i*2],gpuResetSequence[i*2+1]);
 
 	GPUCMD_Finalize();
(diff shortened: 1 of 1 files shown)


smealum:
wow, oops.

jerwuqu:
FYI: This commit broke sf2dlib
https://github.com/xerpi/sf2dlib/issues/5

Steveice10:
Also broke my GPU code... wonder why.

yuriks:
I was looking at this code again yesterday and realized that the command buffer is never flushed before being submitted to the GSP. Anyone feels like giving this a try again, but replacing GPUCMD_Run with GPUCMD_FlushAndRun?

Kfeavel:
Thank you! So thats why sf2dlib wasn't working for me! All it was giving me was a black screen like @jerwuqu said.

PR #145, opened 2015-08-04
Revert "Fix GPU_Reset having wrong GSP command buffer"

smealum:
Reverts smealum/ctrulib#143


aroulin:
Oh, it has already been reverted.

PR #148, opened 2015-08-17
implement gethostbyaddr

Megazig:


libctru/source/services/soc/soc_gethostbyaddr.c
```diff
@@ -0,0 +1,72 @@
+#include "soc_common.h"
+#include <netdb.h>
+
+#define MAX_HOSTENT_RESULTS 16
+static struct hostent SOC_hostent;
+static char           *SOC_hostent_results[MAX_HOSTENT_RESULTS+1];
+static char           *SOC_hostent_alias = NULL;
+
+int h_errno = 0;
```
fincs: `h_errno` is already defined in [soc_gethostbyname](https://github.com/Megazig/ctrulib/blob/master/libctru/source/services/soc/soc_gethostbyname.c#L9). Pinging @mtheall to see how can this be fixed.

Megazig:
as netdb.h is included by soc_gethostbyaddr.c and int h_errno is declared in soc_gethostbyname.c, it could either be removed from soc_gethostbyaddr.c or moved to a better location that is shared between the two

fincs:
Seems sensible.

Megazig:
I will create a commit moving the h_errno to soc_common.c and then redo this commit without h_errno declared here.

PR #149, opened 2015-08-17
Fix gpu example Makefiles for latest picasso

firstEncounter:
A recent commit to picasso changed the parameter format to `picasso -h [header] -o [output] [input]`

Also added a Makefile to graphics/gpu so it gets included in the parent Makefiles.


fincs:
I'm aware of this. The picasso version you're referring to is still under development. The stable version (1.0) works with the current makefiles. Nevertheless, I will merge this PR once the next picasso version becomes stable.

PR #150, opened 2015-08-18
PS service key type and socket functionality

Megazig:


libctru/include/netdb.h
```diff
@@ -22,6 +23,7 @@ extern "C" {
 
 	extern int	h_errno;
 	struct hostent*	gethostbyname(const char *name);
+	struct hostent*	gethostbyaddr(const char *addr, socklen_t len, int type);
```
mtheall: struct hostent*	gethostbyaddr(const **void** *addr, socklen_t len, int type);

PR #153, opened 2015-08-22
Fix the install rule in libctru’s the Makefile

linkmauve:
The current way was depending on tar and bzip2 to first create a binary
package then extract it in the directory pointed by DEVKITPRO.

This commit makes the install rule only copy the relevant files, and
handle correctly the standard DESTDIR variable, so packagers can use it
to make their own packages.

### DIFF
--- a/libctru/Makefile
+++ b/libctru/Makefile
@@ -111,9 +111,9 @@ dist-src:
 
 dist: dist-src dist-bin
 
-install: dist-bin
-	mkdir -p $(DEVKITPRO)/libctru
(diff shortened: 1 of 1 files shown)


mtheall:
Remove the '/' between $(DESTDIR) and $(DEVKITPRO).

$(DESTDIR) should already have a trailing '/', otherwise a blank value will break relative paths.

linkmauve:
Here, fixed.

linkmauve:
Any news on this issue?

fincs:
The changes in this pull request collide with the devkitARM build process. Unfortunately it cannot be accepted.

PR #157, opened 2015-08-27
Fix building on FreeBSD

endrift:
Seriously though why is this piping `bzip2` into `tar` instead of just using `tar xjf`? (I know it's a relic from older dkp days, but still!)


linkmauve:
I also fixed this same issue in an (imo) better way in PR #153.

endrift:
I definitely don't like my fix. But it would be more consistent with the other dka packages.

PR #158, opened 2015-08-27
Protocol numbers and sockopt IP level

Lectem:
Those are based on both RFC 1700 p.7 and the Internet Assigned Numbers Authority (IANA) http://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml


mtheall:
I haven't tried but I don't know if the socket() service call allows a non-zero protocol value. Can you verify one way or the other before merging this?

Lectem:
Those should not be used with socket() (since we don't have SOCK_RAW). However they have to be used for setsockopt and getsockopt as the level argument.
It seems the recognized levels are those listed, and IPV6 (41). 
I do not have the time to dive deeper into what options types are actually allowed though.

Lectem:
This has been updated with various get/setsockopt options.
Note : TCP_NODELAY still has to be determined

mtheall:
I can't say anything about the correctness of the changes. If you've tested these things I would accept the changes as-is, except that two of the files changed don't have newlines ending the file.

Lectem:
Fixed the newlines, everything has been tested with either SOCK_STREAM/SOCK_DGRAM sockets, except for IPPROTO_UDP.
I couldn't find what options that could be passed to getsockopt when using the IPPROTO_UDP level, but if some options could be used, it for sure has value 17.

PR #161, opened 2015-09-03
Add cam:u support and basic documentation.

Steveice10:
Adds support for almost all of cam:u, as much as I know of anyway.

Crappy basic test that writes very first camera frame to "out.bin" (time constraints):
Binaries: https://www.dropbox.com/s/n7zjso8k3934zcc/CameraTest.zip?dl=1
Source: https://gist.github.com/Steveice10/3f2ddfea716a033b1238

### DIFF
--- a/libctru/include/3ds/services/cam.h
+++ b/libctru/include/3ds/services/cam.h
@@ -1,0 +1,570 @@
+/**
+ * @file cam.h
+ * @brief CAM service for using the 3DS's front and back cameras.
+ */
+#pragma once
+#include <3ds/services/y2r.h>
+#include <3ds/types.h>
+
+/**
+ * @brief Camera connection target ports.
+ */
+typedef enum {
+	PORT_NONE = 0x0,
+	PORT_CAM1 = BIT(0),
+	PORT_CAM2 = BIT(1),
+
+	// Port combinations.
+	PORT_BOTH = PORT_CAM1 | PORT_CAM2,
+} CAMU_Port;
+
+/**
+ * @brief Camera combinations.
+ */
+typedef enum {
+	SELECT_NONE = 0x0,
+	SELECT_OUT1 = BIT(0),
+	SELECT_IN1  = BIT(1),
+	SELECT_OUT2 = BIT(2),
+
+	// Camera combinations.
+	SELECT_IN1_OUT1  = SELECT_OUT1 | SELECT_IN1,
+	SELECT_OUT1_OUT2 = SELECT_OUT1 | SELECT_OUT2,
+	SELECT_IN1_OUT2  = SELECT_IN1 | SELECT_OUT2,
+	SELECT_ALL       = SELECT_OUT1 | SELECT_IN1 | SELECT_OUT2,
+} CAMU_CameraSelect;
+
+/**
+ * @brief Camera contexts.
+ */
+typedef enum {
+	CONTEXT_NONE = 0x0,
+	CONTEXT_A    = BIT(0),
+	CONTEXT_B    = BIT(1),
+
+	// Context combinations.
+	CONTEXT_BOTH = CONTEXT_A | CONTEXT_B,
+} CAMU_Context;
+
+/**
+ * @brief Ways to flip the camera image.
+ */
+typedef enum {
+	FLIP_NONE       = 0x0,
+	FLIP_HORIZONTAL = 0x1,
+	FLIP_VERTICAL   = 0x2,
+	FLIP_REVERSE    = 0x3,
+} CAMU_Flip;
+
+/**
+ * @brief Camera image resolutions.
+ */
+typedef enum {
+	SIZE_VGA         = 0x0,
+	SIZE_QVGA        = 0x1,
+	SIZE_QQVGA       = 0x2,
+	SIZE_CIF         = 0x3,
+	SIZE_QCIF        = 0x4,
+	SIZE_DS_LCD      = 0x5,
+	SIZE_DS_LCDx4    = 0x6,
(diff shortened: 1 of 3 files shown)


libctru/source/services/cam.c
```diff
@@ -0,0 +1,737 @@
...
+
+Result CAMU_SetBrightnessSynchronization(bool brightnessSynchronization) {
+    Result ret = 0;
+    u32* cmdbuf = getThreadCommandBuffer();
+    cmdbuf[0] = 0x003E0040;
+    cmdbuf[2] = brightnessSynchronization;
+
+    if ((ret = svcSendSyncRequest(camHandle)) != 0) return ret;
+    return cmdbuf[1];
+}
```
archshift: Newline at EOF

archshift:
If you find time, could you at some point add documentation to 3dbrew about cam:u?

libctru/source/services/cam.c
Steveice10: Gotcha, I'll be making some touch-ups when I get the chance tomorrow. (e.g. Another thing I realized was that I could've just specified bit size for enum fields in structs instead of typing them as u8)

Steveice10:
Sure, I can try to do that when I get the time.

archshift:
All the structs you added which are serialized for IPC need to be packed with `#pragma pack(1)` so the compiler doesn't mistakenly insert any padding anywhere.

yuriks:
@archshift I would change the `bool`s to `u8` to be explicit, but otherwise they're all naturally aligned, so that's not necessary.

Steveice10:
Current issues should be fixed. I added the newline to the end of the file, added "#pragma pack(1)" to the structs, and changed the enum fields from u8 with enum type info in the docs to the enum type with the number of bits specified.

libctru/include/3ds/services/cam.h
```diff
@@ -0,0 +1,809 @@
...
+     * @brief Y axis translation to match the left camera image with the right.
+     */
+    float translationY;
+    /**
+     * @brief X axis rotation to match the left camera image with the right.
+     */
+    float rotationX;
+    /**
+     * @brief Y axis rotation to match the left camera image with the right.
+     */
```
Lectem: If you only give a brief description use "type member;///< description" syntax. It will be way easier to browse it

yuriks: +1

yuriks:
I would not use #pragma pack if it's not necessary (which it isn't here) because it might make the compiler generate more inefficient code, since it changes the struct as a whole to have byte-alignment.

profi200:
Also were got you all the infos from? It"s so detailed and complete that it can't be through reverse engineering or trial and error.

Steveice10:
Documentation issues should be fixed, and pragma pack removed.

The information was gathered and refined over time. It's primarily a mix of RE of functions and their usages, observed behavior, guessing and testing, similarities to Y2R, and identification based on general camera info.

libctru/include/3ds/services/cam.h
```diff
@@ -0,0 +1,648 @@
...
+    CAMU_WhiteBalance whiteBalance : 8;     ///< Camera white balance.
+    s8 sharpness;                           ///< Camera sharpness.
+    bool autoExposureOn;                    ///< Whether to automatically determine the proper exposure.
+    bool autoWhiteBalanceOn;                ///< Whether to automatically determine the white balance mode.
+    CAMU_FrameRate frameRate : 8;           ///< Camera frame rate.
+    CAMU_PhotoMode photoMode : 8;           ///< Camera photo mode.
+    CAMU_Contrast contrast : 8;             ///< Camera contrast.
+    CAMU_LensCorrection lensCorrection : 8; ///< Camera lens correction.
+    bool noiseFilterOn;                     ///< Whether to enable the camera's noise filter.
+    int : 8;                                ///< Padding.
```
Lectem: u8 would be better here

fincs: Those `bool` fields should be `u8` as well (assuming that's the space they take).

Steveice10: I can go ahead and change the padding to a u8. For the bools, which is more preferred, typing as u8 for size, or typing as bool and using the same syntax as the enum fields to preserve type meaning while ensuring size?

Lectem: Well, thing is, while I like bitfields since it can ensure type safety, the problem with those is that its up to the compiler to (re)arrange them in any order. While GCC usually does what you could expect from it when using byte aligned bitfields, and that we do use those in a few other places, I'd rather use a u8 instead of bool+bitfield.
I also think you should check those structures size and check if the members are indeed in the order you expect them to be if they are to be passed to services commands.

yuriks: @Lectem: Bitfield ordering is implementation-specified and is part of the ABI, so in practice, once you get it right once, it's never going to change.

Steveice10:
Struct types fixed to match field size, struct order and size double checked against command buffer writes.

libctru/include/3ds/services/cam.h
Lectem: Well yes, but I was thinking more of the bool than the enums, my point was to make sure those fields were in the good order and that the bool did have the good size

profi200:
Did you use that bowling game to get some infos about the cam service?

Steveice10:
That is one source I used, yes. There were a few camera functions there, but not too many.

fincs:
Could you recommit this using tabs instead of spaces for indentation?

Steveice10:
Done.

PR #162, opened 2015-09-05
GPU: Clean up float conversion functions

yuriks:
The f31 one specially was a clusterfuck. The new functions also handle overflow of the exponent correctly.


yuriks:
(Btw, these have been brute-force verified to match the old functions for all possible inputs, before the overflow fix.)

PR #164, opened 2015-09-08
IPC stuff and DSP service commands

Lectem:
The IPC helpers were added for the sake of clarity and removal of magic numbers.
This isn't the full list of the DSP service commands, however it already covers more than what games seems to be using.

### DIFF
--- a/libctru/include/3ds/ipc.h
+++ b/libctru/include/3ds/ipc.h
@@ -1,0 +1,101 @@
+/**
+ * @file ipc.h
+ * @brief Inter Process Communication helpers
+ */
+
+#pragma once
+
+#include <3ds/types.h>
+
+typedef enum
+{
+	IPC_BUFFER_R = 0x2,
+	IPC_BUFFER_W = 0x4,
+} IPC_BufferRights;
+
+
+/**
+ * @brief Command header to be used for IPC
+ * @param normal_params    Number of normal parameters. Up to 63.
+ * @param translate_params Number of translate parameters. Up to 63.
+ */
+static inline u32 IPC_MakeHeader(u16 command_id, unsigned normal_params, unsigned translate_params)
+{
+	return ((u32) command_id << 16) | (((u32) normal_params & 0x3F) << 6) | (((u32) translate_params & 0x3F) << 0);
+}
+
+
+/**
+ * @brief Creates the header to share handles
+ * @param number The number of handles following this header. Max 64.
+ *
+ * The #number next values are handles that will be shared between the two processes.
+ *
+ * @note Zero values will have no effect.
+ */
+static inline u32 IPC_Desc_SharedHandles(unsigned number)
+{
+	return ((u32)(number - 1) << 26);
+}
+
+/**
+ * @brief Creates the header to transfer handle ownership
+ * @param number The number of handles following this header. Max 64.
+ *
+ * The #number next values are handles that will be duplicated and closed by the other process.
+ *
+ * @note Zero values will have no effect.
+ */
+static inline u32 IPC_Desc_MoveHandles(unsigned number)
+{
+	return ((u32)(number - 1) << 26) | 0x10;
+}
+
+/**
+ * @brief Asks the kernel to fill the handle with the current process handle.
+ *
+ * The next value is a placeholder that will be replaced by the current process handle by the kernel.
+ */
+static inline u32 IPC_Desc_CurProcessHandle(void)
+{
+	return 0x20;
+}
+
+/**
+ * @brief Creates a header describing a static buffer.
+ * @param size      Size of the buffer. Max ?0x03FFFF?.
+ * @param buffer_id The Id of the buffer. Max 0xF.
+ *
+ * The next value is a pointer to the buffer. It will be copied to TLS offset 0x180 + static_buffer_id*8.
+ */
+static inline u32 IPC_Desc_StaticBuffer(size_t size, unsigned buffer_id)
+{
+	return (size << 14) | ((buffer_id & 0xF) << 10) | 0x2;
+}
+
+/**
+ * @brief Creates a header describing a buffer to be sent over PXI.
+ * @param size         Size of the buffer. Max 0x00FFFFFF.
+ * @param buffer_id    The Id of the buffer. Max 0xF.
+ * @param is_read_only true if the buffer is read-only. If false, the buffer is considered to have read-write access.
+ *
+ * The next value is a phys-address of a table located in the BASE memregion.
+ */
+static inline u32 IPC_Desc_PXIBuffer(size_t size, unsigned buffer_id,bool is_read_only)
+{
+	u8 type = 0x4;
+	if(is_read_only)type = 0x6;
+	return (size << 8) | ((buffer_id & 0xF) << 4) | type;
+}
+
+/**
+ * @brief Creates a header describing a buffer from the main memory.
+ * @param size   Size of the buffer. Max 0x0FFFFFFF.
+ * @param rights The rights of the buffer for the destination process
+ *
+ * The next value is a pointer to the buffer.
+ */
+static inline u32 IPC_Desc_Buffer(size_t size, IPC_BufferRights rights)
+{
+	return (size << 4) | 0x8 | rights;
+}
--- a/libctru/source/services/dsp.c
+++ b/libctru/source/services/dsp.c
@@ -1,0 +1,254 @@
+#include <3ds/types.h>
+#include <3ds/svc.h>
+#include <3ds/srv.h>
+#include <3ds/ipc.h>
+#include <3ds/services/dsp.h>
+
+Handle dspHandle = 0;
+
+
+Result dspInit(void)
+{
+	Result ret = 0;
+
+	if (dspHandle == 0)
+	{
+		ret = srvGetServiceHandle(&dspHandle, "dsp::DSP");
+		if (ret < 0) return ret;
+	}
+	if (ret < 0) return ret;
+	DSP_UnloadComponent();
+	return 0;
+}
+
+Result dspExit(void)
+{
+	Result ret = 0;
+//No need to call unload, it will be done automatically by closing the handle
+	if (dspHandle != 0)
+	{
+		ret = svcCloseHandle(dspHandle);
+		if (ret < 0) return ret;
+		dspHandle = 0;
+	}
+
+	return 0;
+}
+
+
+Result DSP_GetHeadphoneStatus(bool* is_inserted)
+{
+	Result ret = 0;
+	u32* cmdbuf = getThreadCommandBuffer();
+	cmdbuf[0] = IPC_MakeHeader(0x001F,0,0);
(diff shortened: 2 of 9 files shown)


libctru/source/services/dsp.c
```diff
@@ -0,0 +1,254 @@
+#include <3ds/types.h>
+#include <3ds/svc.h>
+#include <3ds/srv.h>
+#include <3ds/ipc.h>
+#include <3ds/services/dsp.h>
+
+Handle dspHandle = 0;
```
mtheall: Should be static

libctru/include/3ds/ipc.h
```diff
@@ -0,0 +1,101 @@
...
+ */
+
+#pragma once
+
+#include <3ds/types.h>
+
+typedef enum
+{
+	IPC_BUFFER_R = 0x2,
+	IPC_BUFFER_W = 0x4,
```
mtheall: Prefer (1 << x) style for flags, e.g.:
```
IPC_BUFFER_R = (1 << 1),
IPC_BUFFER_W = (1 << 2),
```

libctru/source/services/dsp.c
```diff
@@ -0,0 +1,254 @@
...
+Result dspInit(void)
+{
+	Result ret = 0;
+
+	if (dspHandle == 0)
+	{
+		ret = srvGetServiceHandle(&dspHandle, "dsp::DSP");
+		if (ret < 0) return ret;
+	}
+	if (ret < 0) return ret;
```
mtheall: `ret >= 0` is guaranteed here, so this line is a no-op.

libctru/source/services/dsp.c
```diff
@@ -0,0 +1,254 @@
...
+
+	return 0;
+}
+
+
+Result DSP_GetHeadphoneStatus(bool* is_inserted)
+{
+	Result ret = 0;
+	u32* cmdbuf = getThreadCommandBuffer();
+	cmdbuf[0] = IPC_MakeHeader(0x001F,0,0);
```
mtheall: The other wrappers here just use `0xHH` (no leading 0's), so to be consistent I would use `0x1F`.

libctru/include/3ds/ipc.h
mtheall: Is it possible to have both R and W? If so I would add:
```
IPC_BUFFER_RW = IPC_BUFFER_R | IPC_BUFFER_W,
```

libctru/source/services/dsp.c
```diff
@@ -0,0 +1,254 @@
...
+	u32* cmdbuf = getThreadCommandBuffer();
+	cmdbuf[0] = IPC_MakeHeader(0x10,3,0);
+	cmdbuf[1] = channel;
+	cmdbuf[2] = unk1;
+	cmdbuf[3] = length;
+
+	u32 * staticbufs = cmdbuf + 0x100;
+
+	u32 saved1 = staticbufs[0x0];
+	u32 saved2 = staticbufs[0x4];
```
Lectem: should be staticbufs[1]

libctru/include/3ds/ipc.h
Lectem: Yes it is, I will add it

libctru/source/services/dsp.c
```diff
@@ -0,0 +1,254 @@
...
+}
+
+
+Result DSP_GetHeadphoneStatus(bool* is_inserted)
+{
+	Result ret = 0;
+	u32* cmdbuf = getThreadCommandBuffer();
+	cmdbuf[0] = IPC_MakeHeader(0x001F,0,0);
+	if ((ret = svcSendSyncRequest(dspHandle)) != 0) return ret;
+	*is_inserted = cmdbuf[2] & 0xFF;
```
mtheall: It would be nice to make sure *is_inserted can only be output as `0` or `1`.
```
*is_inserted = (cmdbuf[2] & 0xFF) ? 1 : 0;
```

libctru/source/services/dsp.c
```diff
@@ -0,0 +1,254 @@
...
+	Result ret = 0;
+	u32* cmdbuf = getThreadCommandBuffer();
+	cmdbuf[0] = IPC_MakeHeader(0x11,3,2);
+	cmdbuf[1] = size;
+	cmdbuf[2] = prog_mask;
+	cmdbuf[3] = data_mask;
+	cmdbuf[4] = IPC_Desc_Buffer(size,IPC_BUFFER_R);
+	cmdbuf[5] = (u32) component;
+	if ((ret = svcSendSyncRequest(dspHandle)) != 0) return ret;
+	*is_loaded = cmdbuf[2] & 0xFF;
```
mtheall: It would be nice to make sure `*is_inserted` can only be output as `0` or `1`.
```
*is_loaded = (cmdbuf[2] & 0xFF) ? 1 : 0;
```

libctru/source/services/dsp.c
```diff
@@ -0,0 +1,254 @@
...
+	cmdbuf[1] = channel;
+	cmdbuf[2] = unk1;
+	cmdbuf[3] = length;
+
+	u32 * staticbufs = cmdbuf + 0x100;
+
+	u32 saved1 = staticbufs[0x0];
+	u32 saved2 = staticbufs[0x4];
+
+	staticbufs[0] = (length<<14) | 2;
```
mtheall: Should this be:
```
IPC_Desc_StaticBuffer(length, 0);
```

It should give the right value, but I don't know if this is what that function is for.

libctru/source/services/dsp.c
```diff
@@ -0,0 +1,254 @@
...
+	cmdbuf[1] = interrupt;
+	cmdbuf[2] = channel;
+	cmdbuf[3] = IPC_Desc_SharedHandles(1);
+	cmdbuf[4] = handle;
+	if ((ret = svcSendSyncRequest(dspHandle)) != 0) return ret;
+	return cmdbuf[1];
+}
+
+
+Result DSP_ReadPipeIfPossibleEx(u32 channel,u32 unk1, u8 const *buffer, u16 length, u16* length_read)
```
mtheall: Should this be static? The API is not exported in `dsp.h`.

libctru/source/services/dsp.c
```diff
@@ -0,0 +1,254 @@
...
+Result DSP_ReadPipeIfPossible(u32 channel, u8 const *buffer, u16 length, u16* length_read)
+{
+	return DSP_ReadPipeIfPossibleEx(channel,0,buffer,length, length_read);
+}
+
+Result DSP_WriteProcessPipe(u32 channel, u8 const *buffer, u32 length)
+{
+	Result ret = 0;
+	u32* cmdbuf = getThreadCommandBuffer();
+	cmdbuf[0] = IPC_MakeHeader(0xd,2,2);
```
mtheall: Hex values everywhere else are all caps: `0xD`

libctru/source/services/dsp.c
```diff
@@ -0,0 +1,254 @@
...
+	cmdbuf[4] = (u32) buffer;
+	if ((ret = svcSendSyncRequest(dspHandle)) != 0) return ret;
+	return cmdbuf[1];
+}
+
+Result DSP_ConvertProcessAddressFromDspDram(u32 dsp_address, u32 *arm_address)
+{
+	Result ret = 0;
+	u32* cmdbuf = getThreadCommandBuffer();
+	cmdbuf[0] = IPC_MakeHeader(0xc,1,0);
```
mtheall: Hex values everywhere else are all caps: `0xC`

libctru/source/services/dsp.c
Lectem: I'm currently investigating wether it the values in the static buffers areas should be using the buffer id or not

libctru/source/services/dsp.c
```diff
@@ -0,0 +1,254 @@
...
+}
+
+Result DSP_RecvDataIsReady(u16 regNo, bool * is_ready)
+{
+	Result ret = 0;
+	u32* cmdbuf = getThreadCommandBuffer();
+	cmdbuf[0] = IPC_MakeHeader(0x2,1,0);
+	cmdbuf[1] = regNo;
+	if ((ret = svcSendSyncRequest(dspHandle)) != 0) return ret;
+	*is_ready = cmdbuf[2] & 0xFF;
```
mtheall: It would be nice to make sure `*is_ready` can only be output as `0` or `1`.
```
*is_ready = (cmdbuf[2] & 0xFF) ? 1 : 0;
```

libctru/source/services/dsp.c
```diff
@@ -0,0 +1,254 @@
...
+}
+
+Result DSP_SendDataIsEmpty(u16 regNo, bool * is_empty)
+{
+	Result ret = 0;
+	u32* cmdbuf = getThreadCommandBuffer();
+	cmdbuf[0] = IPC_MakeHeader(0x4,1,0);
+	cmdbuf[1] = regNo;
+	if ((ret = svcSendSyncRequest(dspHandle)) != 0) return ret;
+	*is_empty = cmdbuf[2] & 0xFF;
```
mtheall: It would be nice to make sure `*is_empty` can only be output as `0` or `1`.
```
*is_empty = (cmdbuf[2] & 0xFF) ? 1 : 0;
```

libctru/source/services/dsp.c
mtheall: Yes, I was just going to comment that I thought this was weird. I assume the 0x4 is a translation error from u8-based indexing.

libctru/source/services/dsp.c
Lectem: I didn't expose it because from what I saw, unk1 is always set to 0.
The plan was to have DSP_ReadPipeIfPossible signature change and replace DSP_ReadPipeIfPossibleEx if we ever find use for this unknown parameter

libctru/source/services/dsp.c
Lectem: It is, will fix it soon

libctru/source/services/dsp.c
Lectem: Ok so I found the name of the argument from BPB, so I'll remove DSP_ReadPipeIfPossibleEx

libctru/source/services/dsp.c
Lectem: bool is a define for _Bool, and the standard says "When any scalar value is converted to _Bool, the result is 0 if the value compares equal to 0; otherwise, the result is 1." so I assume that it is not needed ?

mtheall: Please give a specific standard when citing it. As far as I can tell, this comes from C99. In the interest of portability, I would prefer to explicitly set the value to `0` or `1`. I would alternatively accept `#include <stdbool.h>` which would require this file be compiled with C99 or newer and enforces the conversion to `0` or `1` as you said.

Lectem: stdbool.h is already included from https://github.com/smealum/ctrulib/blob/master/libctru/include/3ds/types.h#L8, that's why I didn't bother mentionning it

PR #166, opened 2015-09-09
IPC refactor

Lectem:
I'm opening this pull request so that people can start reviewing it and give their opinion.
Please do not merge it now since I will be adding/modifying commits for other services later.


Lectem:
Refactor complete !

PR #169, opened 2015-09-15
GSP: Use atomics in interrupt reading loop

yuriks:
I wrote this patch 9 months ago in January (cc @plutooo) and now not finishing and PRing it finally came to bite me in the ass. :)

This appears to fix the rare random crashes that were happening in mGBA (cc @endrift) and probably also in Portal3DS (cc @smealum) and blargSnes (cc @StapleButter) where GSP interrupts would randomly be lost, causing the application to hang forever unless a timeout was added to all waits.

### DIFF
--- a/libctru/source/services/gsp.c
+++ b/libctru/source/services/gsp.c
@@ -82,6 +82,50 @@ void gspWaitForEvent(GSP_Event id, bool nextEvent)
 		svcClearEvent(gspEvents[id]);
 }
 
+static int popInterrupt()
+{
+	int curEvt;
+	u32 strexFailed;
+	do {
+		union {
+			struct {
+				u8 cur;
+				u8 count;
+				u8 err;
+				u8 unused;
+			};
+			u32 as_u32;
+		} header;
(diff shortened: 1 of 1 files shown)


endrift:
I spent a good chunk of the day testing an earlier version of this patch, so I can verify that it fixes the freezes.

smealum:
well that's awesome. thanks for putting all this work into it !

mtheall:
Do gcc atomic builtins not work? I would prefer those over inline asm.

yuriks:
@mtheall The operation inside the critical section is quite complex, so no, I don't think you can use the builtins. (I would have preferred to use those too. Inline asm for ldrex/strex is kinda nerve-wracking.)

mtheall:
Also, the use of this union is illegal. You are not allowed to read from a union member if that member was not the last one written. The only way I can see this done legally is with bit shifting the header fields.

neobrain:
I guess this might have been an issue for citrace-player, too.

PR #170, opened 2015-09-19
document ArbitrateAddress + update ArbitrationType

Lectem:
This has been tested on n3ds 9.9.

### DIFF
--- a/libctru/include/3ds/svc.h
+++ b/libctru/include/3ds/svc.h
@@ -73,12 +73,15 @@ typedef struct {
 } PageInfo;
 
 typedef enum {
-	ARBITER_FREE            = 0,
-	ARBITER_ACQUIRE         = 1,
-	ARBITER_KERNEL2         = 2,
-	ARBITER_ACQUIRE_TIMEOUT = 3,
-	ARBITER_KERNEL4         = 4,
+	ARBITRATION_SIGNAL                                  = 0, ///< Signal #value threads for wake-up.
+	ARBITRATION_WAIT_IF_LESS_THAN                       = 1, ///< If the memory at the address is strictly lower than #value, then wait for signal.
+	ARBITRATION_DECREMENT_AND_WAIT_IF_LESS_THAN         = 2, ///< If the memory at the address is strictly lower than #value, then decrement it and wait for signal.
+	ARBITRATION_WAIT_IF_LESS_THAN_TIMEOUT               = 3, ///< If the memory at the address is strictly lower than #value, then wait for signal or timeout.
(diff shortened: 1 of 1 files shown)


mtheall:
We should implement semaphores and condition variables using this synchronization primitive so users understand how the heck to use this thing.

Lectem:
@yuriks was working on a lightweight mutex/critical section, but a condition variable implementation would be interesting too I suppose

libctru/include/3ds/svc.h
```diff
@@ -73,13 +73,16 @@ typedef struct {
...
 
 typedef enum {
-	ARBITER_FREE            = 0,
-	ARBITER_ACQUIRE         = 1,
-	ARBITER_KERNEL2         = 2,
-	ARBITER_ACQUIRE_TIMEOUT = 3,
-	ARBITER_KERNEL4         = 4,
+	ARBITRATION_SIGNAL                                  = 0, ///< Signal #value threads for wake-up.
+	ARBITRATION_WAIT_IF_LESS_THAN                       = 1, ///< If the memory at the address is strictly lower than #value, then wait for signal.
+	ARBITRATION_DECREMENT_AND_WAIT_IF_LESS_THAN         = 2, ///< If the memory at the address is strictly lower than #value, then decrement it and wait for signal.
```
yuriks: I think the comment is wrong here. Isn't the decrement done *before* the check?

yuriks:
@mtheall Indeed, but I don't know anyone with expertise in implementing concurrency primitives, and I seem to suck at this.

libctru/include/3ds/svc.h
Lectem: No, I tried it with value 1 and the memory containing 1, it would never wait, while it would work if the value was 0 or somerhing else

yuriks: Did you check what was then the value of the address while the thread waited, and then afterwards after it was signalled?

Lectem: Well I'm using it in a loop, and it kept not waiting so...

Lectem: in order : 
- check the cond
- decrement
- wait for signal
- resume without checking the condition

mtheall:
I have implemented the semaphore and condition variables a while back. I can look for that code and try to fold it in.

Lectem:
@mtheall that would be nice !

Lectem:
@mtheall any news for the semaphore?

mtheall:
Sorry I haven't had a chance to look for it yet. I'll submit a PR and reference here when I can get around to it!

PR #176, opened 2015-09-30
Implement some services from gsp::Lcd.

thedax:
http://www.3dbrew.org/wiki/GSP_Services

~~Kind of like svcBackdoor, this would require service access check patching (and therefore ARM11 kernel, I suppose?), but it'd be nice to have it for power saving.~~ It seems my makerom from project_ctr was outdated, and it's working now without requiring service access patching. I can't get my test program to work in ninjhax2 though.

One word of warning when using this service: make sure to release the handle before letting the home button be pressed/letting your application be interrupted by something; the entire 3DS console seems to hang if the currently running program has the handle. It's also a good idea to turn whichever backlight(s) on before exiting the application, since the home menu doesn't seem to re-enable it (not sure if the Homebrew Launcher does, but I assume it doesn't).

Thanks to @Megazig, @Normmatt and the others who helped me on IRC the other night.

### DIFF
--- a/libctru/include/3ds/services/gsp.h
+++ b/libctru/include/3ds/services/gsp.h
@@ -50,9 +50,9 @@ typedef enum
 
 typedef enum
 {
-	GSPLCD_TOP=1,
-	GSPLCD_BOTTOM=2,
-	GSPLCD_BOTH=3,
+	GSPLCD_TOP    = BIT(0),
+	GSPLCD_BOTTOM = BIT(1),
+	GSPLCD_BOTH   = GSPLCD_TOP | GSPLCD_BOTTOM,
 }GSPLCD_Screens;
 
 Result gspInit();
@@ -90,6 +90,5 @@ Result GSPGPU_UnregisterInterruptRelayQueue(Handle* handle);
 Result GSPGPU_TriggerCmdReqQueue(Handle *handle);
 Result GSPGPU_SubmitGxCommand(u32* sharedGspCmdBuf, u32 gxCommand[0x8], Handle* handle);
 
-// 1 = top, 2 = bottom, 3 = both
-Result GSPLCD_PowerOffBacklight(u32 screen);
-Result GSPLCD_PowerOnBacklight(u32 screen);
+Result GSPLCD_PowerOffBacklight(GSPLCD_Screens screen);
+Result GSPLCD_PowerOnBacklight(GSPLCD_Screens screen);
--- a/libctru/include/3ds/services/gsp.h
+++ b/libctru/include/3ds/services/gsp.h
@@ -48,6 +48,13 @@ typedef enum
 	GSPEVENT_MAX, // used to know how many events there are
 } GSP_Event;
 
+typedef enum
+{
+	GSPLCD_TOP=1,
+	GSPLCD_BOTTOM=2,
+	GSPLCD_BOTH=3,
+}GSPLCD_Screens;
+
 Result gspInit();
 void gspExit();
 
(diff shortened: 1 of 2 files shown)


smealum:
we can actually get gsp::Lcd in ninjhax2/tubehax/ironhax/whatever. is it
really that useful ? what for ?

smealum:
oh is it just for the backlight ?

thedax:
Yeah, I only added backlight commands.

thedax:
It would be most useful for applications like RetroArch that don't use the bottom screen for much other than an info console, or homebrew that don't use the bottom screen at all.

thedax:
Oops, I have a typo in one of the variable names. Fixing.

Normmatt:
I'd say use an enum for the params something like Backlight_Top, Backlight_Bottom, Backlight_Both as using 0,1,2 isn't very nice.

thedax:
Sure, I can add it.

thedax:
Added. If the names aren't desirable, please tell me how I can improve them.

libctru/include/3ds/services/gsp.h
```diff
@@ -48,9 +48,19 @@ typedef enum
 	GSPEVENT_MAX, // used to know how many events there are
 } GSP_Event;
 
+typedef enum
+{
+	GSPLCD_TOP=1,
+	GSPLCD_BOTTOM=2,
+	GSPLCD_BOTH=3,
+}GSPLCD_Screens;
```
Lectem: Could you please change this to the following ?

```
typedef enum
{
	GSPLCD_TOP    = BIT(1),
	GSPLCD_BOTTOM = BIT(2),
	GSPLCD_BOTH   = GSPLCD_TOP | GSPLCD_BOTTOM,
}GSPLCD_Screens;
```

libctru/include/3ds/services/gsp.h
```diff
@@ -79,3 +89,7 @@ Result GSPGPU_RegisterInterruptRelayQueue(Handle *handle, Handle eventHandle, u3
 Result GSPGPU_UnregisterInterruptRelayQueue(Handle* handle);
 Result GSPGPU_TriggerCmdReqQueue(Handle *handle);
 Result GSPGPU_SubmitGxCommand(u32* sharedGspCmdBuf, u32 gxCommand[0x8], Handle* handle);
+
+// 1 = top, 2 = bottom, 3 = both
+Result GSPLCD_PowerOffBacklight(u32 screen);
+Result GSPLCD_PowerOnBacklight(u32 screen);
```
Lectem: u32 can safely be replaced by GSPLCD_Screens here, and comment removed

libctru/include/3ds/services/gsp.h
plutooo: BIT(0) you mean?

Lectem: Whoops sorry, yes of course. I'll update the note

thedax:
> we can actually get gsp::Lcd in ninjhax2/tubehax/ironhax/whatever.

@smealum, is this something the end-user can test right now with an xml file or something (like the Wii, which can ask for AHBPROT, for example), or is it something on your end that has to be added into ninjhax?

smealum:
nothing for the user to do, it will come with next release

thedax:
Okay, thanks.

smealum:
alright can confirm that ninjhax2.5 (and by extension tubehax, ironhax, browserhax, themehax, oot3dhax...) will have access to gsp::Lcd without any trouble, just tested a demo from @Megazig and it worked fine. let me know if you have any trouble with it once it's out.

Megazig:
for the svcSyncRequest calls, you should check if ret and return if not equal to 0, otherwise return cmdbuf[1];

thedax:
Done. I tried to make it match nearby coding styles.

Megazig:
LGTM

PR #177, opened 2015-10-01
Implement APT_GetProgramID

suloku:


libctru/source/services/apt.c
```diff
@@ -833,6 +833,28 @@ Result APT_GetAppletProgramInfo(Handle* handle, u32 id, u32 flags, u16 *titlever
...
+	cmdbuf[1] = 0x20;
+	
+	Result ret=0;
+	if((ret=svcSendSyncRequest(*handle)))return ret;
+
+	if(ret==0)ret = cmdbuf[1];
+
+	if(pProgramID)
+	{
+		*pProgramID = 0;
```
profi200: That is not needed.

PR #180, opened 2015-10-03
Dynamically allocate shared memory.

Steveice10:
This PR adds an allocator for the mappable memory region (0x10000000 - 0x14000000) and switches the default shared memory blocks of GFX, CSND, IRRST, and HID to this system.

Better error handling was also added to csndInit to properly close handles and free memory on failure.

### DIFF
--- a/libctru/source/allocator/mappable.cpp
+++ b/libctru/source/allocator/mappable.cpp
@@ -1,0 +1,61 @@
+extern "C"
+{
+	#include <3ds/types.h>
+	#include <3ds/mappable.h>
+	#include <3ds/util/rbtree.h>
+}
+
+#include "mem_pool.h"
+#include "addrmap.h"
+
+static MemPool sMappablePool;
+
+static bool mappableInit()
+{
+	auto blk = MemBlock::Create((u8*)0x10000000, 0x04000000);
+	if (blk)
+	{
+		sMappablePool.AddBlock(blk);
+		rbtree_init(&sAddrMap, addrMapNodeComparator);
+		return true;
+	}
+	return false;
+}
+
+void* mappableAlloc(size_t size)
+{
+	// Initialize the pool if it is not ready
+	if (!sMappablePool.Ready() && !mappableInit())
+		return nullptr;
+
+	// Allocate the chunk
+	MemChunk chunk;
+	if (!sMappablePool.Allocate(chunk, size, 12))
+		return nullptr;
+
+	auto node = newNode(chunk);
+	if (!node)
+	{
+		sMappablePool.Deallocate(chunk);
+		return nullptr;
+	}
+	if (rbtree_insert(&sAddrMap, &node->node));
+	return chunk.addr;
+}
+
+void mappableFree(void* mem)
+{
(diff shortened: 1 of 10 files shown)


libctru/source/allocator/mappable.cpp
```diff
@@ -0,0 +1,74 @@
...
+	if (shift == 32) // Invalid alignment
+		return nullptr;
+
+	// Initialize the pool if it is not ready
+	if (!sMappablePool.Ready() && !mappableInit())
+		return nullptr;
+
+	// Allocate the chunk
+	MemChunk chunk;
+	if (!sMappablePool.Allocate(chunk, size, shift))
```
fincs: This could have been simplified to `if (!sMappablePool.Allocate(chunk, size, 12)) // Force page alignment`, and the above section that calculates the shift removed.

libctru/source/gfx.c
```diff
@@ -5,12 +5,13 @@
...
 #include <3ds/svc.h>
 #include <3ds/linear.h>
+#include <3ds/mappable.h>
 #include <3ds/vram.h>
 
 GSP_FramebufferInfo topFramebufferInfo, bottomFramebufferInfo;
 
 u8 gfxThreadID;
-u8* gfxSharedMemory;
+u8* gfxSharedMemory = NULL;
```
fincs: Setting it to NULL is not necessary since the variable lies in the `.bss` region which is set to zero.

libctru/source/services/hid.c
```diff
@@ -42,6 +42,18 @@ Result hidInit(u32* sharedMem)
 
 	// Map HID shared memory at addr "sharedMem".
 	hidSharedMem=sharedMem;
+	if(!hidSharedMem)
```
fincs: hidInit's sharedMem parameter should have been removed, instead of preserving it and managing ownership.

libctru/source/services/irrst.c
```diff
@@ -39,6 +40,18 @@ Result irrstInit(u32* sharedMem)
 
 	// Map ir:rst shared memory at addr "sharedMem".
 	irrstSharedMem=sharedMem;
+	if(!irrstSharedMem)
```
fincs: irrstInit's sharedMem parameter should have been removed, instead of preserving it and managing ownership.

libctru/source/system/allocateHeaps.c
```diff
@@ -17,8 +19,13 @@ void __attribute__((weak)) __system_allocateHeaps() {
 
 	// Allocate the linear heap
 	svcControlMemory(&__linear_heap, 0x0, 0x0, __linear_heap_size, MEMOP_ALLOC_LINEAR, MEMPERM_READ | MEMPERM_WRITE);
+
+	// Set up the mappable heap
+	__mappable_heap = 0x10000000;
```
fincs: I think these variables are unnecessary. It's best to hardcode this in mappableInit.

Steveice10:
Issues should be resolved.

fincs:
LGTM now.

PR #189, opened 2015-10-07
fixed enums for c++

Lectem:
Fixed problems with enums, those were originally anonymous, and no typedef was added to name them, introducing linker problems later on with c++ projects.

### DIFF
--- a/libctru/include/3ds/services/apt.h
+++ b/libctru/include/3ds/services/apt.h
@@ -64,7 +64,7 @@ enum {
 	APTSIGNAL_UTILITY      = 9,  ///< Utility called.
 	APTSIGNAL_SLEEPSYSTEM  = 10, ///< System sleeping.
 	APTSIGNAL_ERROR        = 11  ///< Error occurred.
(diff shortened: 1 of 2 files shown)


mtheall:
Wouldn't it have been better to do a typedef declaration? It could then retain the name.

fincs:
The names weren't originally there, they were erroneously added by a recent pull request.

Steveice10:
IMO there's no reason for them to not have a name. Helps describe it and allows specific variable typing.

fincs:
IMO there's no reason for them to have a name. C/C++ enumerations are quite broken in general due to their lack of strong typing and the lack of an explicit underlying type (however, C++ `class enum` goes too far and is too strict), and it also does not make semantic sense for bitmask bits to be a named *enumeration*.

profi200:
I never had problems like this. I'm the only one? All my C++ projects compiled fine before with the ctrulib from master.

fincs:
The breakage has only affected great-refactor temporarily for just a few days.

PR #190, opened 2015-10-07
Fix NDSP enums for C++.

Sahnvour:
Continuation of #189


mtheall:
Wouldn't it have been better to do a typedef declaration? It could then retain the name.

fincs:
The names weren't originally there, they were erroneously added by a recent pull request.

PR #196, opened 2015-10-22
Add camera example

wchill:
Currently ctrulib doesn't have any example code for using the camera. This example code takes a 3D picture using the front cameras and displays it on the top screen.

A couple of notes:
* the camera calibration data isn't used yet - so the images will have crappy 3D. Still need to figure this out
* about 25% of the time the code hangs on `CAMU_SetReceiving` (`main.c:89`). On my N3DS, when this happens I will hear the camera shutter sound when I hold the power button to try and power off.
* `camera.xml` is needed when using a userland exploit, or else you will get access denied errors when accessing the camera. `camera.xml` tells the Homebrew Launcher to use the system 3DS Camera app as the entrypoint.

### DIFF
--- a/examples/camera/Makefile
+++ b/examples/camera/Makefile
@@ -1,0 +1,172 @@
+#---------------------------------------------------------------------------------
+.SUFFIXES:
+#---------------------------------------------------------------------------------
+
+ifeq ($(strip $(DEVKITARM)),)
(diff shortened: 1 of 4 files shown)


wchill:
Oops, don't merge yet. Just found a small bug.

PR #203, opened 2015-11-06
expand the original stack to __stacksize__

aliaspider:
instead of moving sp to a different buffer.

I made this PR to the great-refactor branch since it requires a recent 2.5 payload (31-10-2015 or newer) to work correctly for 3dsx builds.


plutooo:
I'm sorry; I really appreciate your work. Best, plutoo

PR #207, opened 2015-11-08
Dynamically determine heap sizes.

Steveice10:
The total amount of memory available is now determined by subtracting the amount of APPLICATION memory used from the value of APPMEMALLOC. This number is then split in half and aligned to page size, giving each heap half of the total available memory.

Let me know if there are any needed changes. (proportion to split memory by, etc)

### DIFF
--- a/libctru/include/3ds/os.h
+++ b/libctru/include/3ds/os.h
@@ -17,7 +17,16 @@
 /// Retrieves the revision version from a packed system version.
 #define GET_VERSION_REVISION(version) (((version)>> 8)&0xFF)
 
-/*! OS_VersionBin. Format of the system version: "<major>.<minor>.<build>-<nupver><region>" */
+/// Memory regions.
+typedef enum
+{
+	MEMREGION_ALL = 0,         ///< All regions.
+	MEMREGION_APPLICATION = 1, ///< APPLICATION memory.
+	MEMREGION_SYSTEM = 2,      ///< SYSTEM memory.
+	MEMREGION_BASE = 3,        ///< BASE memory.
+} MemRegion;
+
+/// OS_VersionBin. Format of the system version: "<major>.<minor>.<build>-<nupver><region>"
 typedef struct
 {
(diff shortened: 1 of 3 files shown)


smealum:
i'm against this. the way things currently work is that the launcher is responsible for telling the app how much memory it can use in the form of the 3dsx header regular and linear heap sizes.

as of hax 2.5, those values are already the maximum possible ones.

what i'd rather see is for the allocator to take the sum of those values and the recalculate the sizes according to some proportion variable which is user configurable (as a weak overridable variable for example).

fincs:
I agree with smealum.

Steveice10:
@smealum @fincs I can see how it wouldn't be necessary from a *hax standpoint, as it already gives it the maximum available memory, but from other environments, such as plain CIA, there would be no mechanism in place to make use of the maximum amount of memory made available by the RSF.

smealum:
yeah that is a good point. so a couple things there :

1. shouldn't assume it's going to be an app. what if it's running as an applet ?
2. it should be overridable by the launcher somehow

Steveice10:
I assume there's some variable within ctrulib that wouldn't be there without a launcher, correct (__service_ptr, if I remember correctly, comes to mind)? If so, I could update it to:

* Check if __service_ptr if set. If it is, use the old size values.
* If not, query the method used in this PR. (although there's still the issue of app vs applet)
* Either way, add both values and distribute based on a defined proportion.

fincs:
There's a reason `__system_allocateHeaps` is weak and in a separate source code file. libctru logic generally gives priority to homebrew loaders instead of CIA. I'm not sure of the benefits of making the default implementation considerably more complex to favor a method of execution that is basically not really recommended anymore.

`__service_ptr` can indeed be checked to detect a homebrew loader.

Steveice10:
Update to use launcher information when provided and split based on modifiable multiplier variables.

Steveice10:
Updated to latest commit.

PR #212, opened 2015-11-08
Add news:s support.

Steveice10:
Adds basic news:s support to the ctrulib news code.


OctopusRift:
Thanks for fixing my issue so quick!

PR #214, opened 2015-11-09
Pretty sure it's not a good idea to decrement a handle.

thedax:
This broke New 3DS speedup, etc..


smealum:
good thing SOMEONE's around to keep fincs's destructive changes in check

fincs:
lel
Thanks for spotting this one.

thedax:
@fincs: You're welcome. :P

PR #216, opened 2015-11-11
Service cleanup.

Steveice10:
Cleans up bits of service code here and there, along with completing the SRV functions.


Lectem:
I dont really think renaming modules that are only composed of one service really need a name change (eg: Y2R is only made of y2r:u, I dont think changing the prefix to Y2RU was needed).

Steveice10:
Technically Y2R is part of the CAM module, but it was mainly for consistency's sake.

Lectem:
While I do get the point of having the same convention for all services, I still think that those that do not need any differentiation such as Y2R and MVD shouldn't use such long prefixes. This is of course only my opinion but MVDSTD_ really is ugly (mvdstd wasn't better).

libctru/include/3ds/services/ac.h
```diff
@@ -10,11 +10,11 @@ Result acInit(void);
 /// Exits AC.
 void acExit(void);
 
+/// Waits for the system to connect to the internet.
+Result acWaitInternetConnection(void);
```
Lectem: Shouldn't this be using the uppercase notation since it's not API and just a raw IPC command ?

Steveice10: It actually isn't an IPC command, for some reason it was uppercased. This is what it was before: https://github.com/Steveice10/ctrulib/commit/7e5edd91385f5943899de00f4de5e40bc91d3c9c#diff-8ff965f94c5b943d375ce717582bc9c7L93

libctru/source/services/gspgpu.c
```diff
@@ -353,86 +384,10 @@ Result GSPGPU_SaveVramSysArea(void)
...
-
-	cmdbuf[0] = 0x00110040;
-	cmdbuf[1] = screen;
-
-	Result ret=0;
-	if (R_FAILED(ret = svcSendSyncRequest(gspLcdHandle)))return ret;
-
-	return cmdbuf[1];
-}
+}
```
Lectem: missing endline

libctru/source/services/mvd.c
```diff
@@ -9,27 +9,27 @@
...
-static Result mvdstdipc_Initialize(u32 *buf, u32 bufsize, Handle kprocess)
+static Result MVDSTD_Initialize(u32* buf, u32 bufsize)
 {
 	u32* cmdbuf = getThreadCommandBuffer();
 	cmdbuf[0] = IPC_MakeHeader(0x1,2,2); // 0x10082
 	cmdbuf[1] = (u32)buf;
 	cmdbuf[2] = bufsize;
 	cmdbuf[3] = IPC_Desc_SharedHandles(1);
-	cmdbuf[4] = kprocess;
+	cmdbuf[4] = CUR_PROCESS_HANDLE;
```
Lectem: is CUR_PROCESS_HANDLE really the only value possible here ?

Steveice10: I don't know if it *has* to be, but I assume its similar to others like GSPGPU_FlushDataCache. No idea what purpose any other values would have, nor what they would even do.

libctru/source/services/gspgpu.c
Steveice10: Should be fixed.

libctru/include/3ds/services/ac.h
Lectem: right sorry, went a bit to fast when I saw the R_SUCCEEDED

fincs:
In general lines, this pull is in the right direction. However there are several things that I don't feel comfortable with:

- srv is not a service, therefore the XYZ_ naming convention does not apply. SRV_Xyz() looks ugly and not in line with svcSystemCalls().
- I'd rather not have GSPGPU commands like gspWaitForVBlank() be renamed to gsp**Gpu**WaitForVBlank(), even if it's inconsistent.
- If an enumeration was anonymous, keep it anonymous and **don't** add a name -- there's a reason for it (namely they are part of a bitmask and it never made sense for them to be an independent type). Enumerations that already have a name can be renamed to something more consistent like you did though.
- If a service has several variations that only differ in privilege but not commands (and this access is already handled automatically by libctru), you can just use the name of the service group instead of worrying about there being several variations. So far you haven't done anything with these, which is correct.

Steveice10:
* SRV is not a service, but it is still IPC like the rest, the only difference being a port handle. Should the convention then only apply to services?
* Noted, does this apply to all the GSPGPU lowercase functions or just the macros?
* Should this be standardized? For example the GSPLCD screen bitmasks have a name (https://github.com/smealum/ctrulib/blob/great-refactor/libctru/include/3ds/services/gsp.h#L61), so should I go through and remove improper names? What exactly is the rule to follow for named vs anonymous enums?

fincs:
> Should the convention then only apply to services?

Yes - and that basically means everything except for srv and srvPm.

> does this apply to all the GSPGPU lowercase functions or just the macros?

All GSPGPU lowercase functions.

> Should this be standardized? For example the GSPLCD screen bitmasks have a name (..) so should I go through and remove improper names?

Yes, please. See below:

> What exactly is the rule to follow for named vs anonymous enums?

If it's a bitmask (i.e. bits that are toggled or untoggled) or an enumerated field inside a bitmask; it should be an anonymous enum, since there do not exist (or should not exist) functions (not macros) that take an isolated value from the enum (instead, the function receives the final combined field). In other case, it should be a named enum, and the type should be explicitly used in function definitions.

Steveice10:
Changes should be taken care of, although:

> Yes - and that basically means everything except for srv and srvPm.

SRVPM is actually a service now, does that still apply?

fincs:
Now it looks perfect, save for the fact that ndspEncoding, CSND_Encoding and CSND_LoopMode are all parts of a bigger bitfield and thus should not have a name. Btw, there are some functions like ndspSetOutputMode, ndspSetClippingMode, ndspSurroundSetPos, ndspChnSetInterp, CSND_SetChnRegsPSG and CSND_SetDuty whose signature needs to be updated to use the new enumeration types.

Steveice10:
Removed names, updated types.

fincs:
Thank you. :+1:

Issue #217, opened 2015-11-11
Make sure IPC getters are using mask for short/char/bool values


Lectem:
I saw a few places where values are not masked while nintendo uses STRB or STRH. We should also make sure that all bool types are masked using 0xFF.
Enum values should be masked depending on the smallest value (usually 0xFF or 0xFFFF). While gcc already stores most of those in bytes or halfwords, it is not guaranteed by the standard.

For example I spotted https://github.com/Steveice10/ctrulib/blob/service-cleanup/libctru/source/services/cam.c#L355 but there are probably many other services suffering from this issue.

Steveice10:
Updated https://github.com/smealum/ctrulib/pull/216 to resolve this.

PR #218, opened 2015-11-12
Bring FS up to date.

Steveice10:
Brings the FS code up to date with the latest 3dbrew documentation, cleaning up types and implementing a bunch of new functions.

### DIFF
--- a/libctru/source/romfs_dev.c
+++ b/libctru/source/romfs_dev.c
@@ -137,10 +137,10 @@ Result romfsInit(void)
 		if (units == (size_t)-1) return 3;
 		__utf16path[units] = 0;
 
-		FS_archive arch = { ARCH_SDMC, { PATH_EMPTY, 1, (u8*)"" }, 0, 0 };
-		FS_path path = { PATH_WCHAR, (units+1)*2, (u8*)__utf16path };
+		FS_Archive arch = { ARCHIVE_SDMC, { PATH_EMPTY, 1, (u8*)"" }, 0 };
+		FS_Path path = { PATH_UTF16, (units+1)*2, (u8*)__utf16path };
 
-		Result rc = FSUSER_OpenFileDirectly(&romFS_file, arch, path, FS_OPEN_READ, FS_ATTRIBUTE_NONE);
+		Result rc = FSUSER_OpenFileDirectly(&romFS_file, arch, path, FS_OPEN_READ, 0);
 		if (R_FAILED(rc)) return rc;
 
 		_3DSX_Header hdr;
@@ -155,10 +155,10 @@ Result romfsInit(void)
 		u8 zeros[0xC];
 		memset(zeros, 0, sizeof(zeros));
 
-		FS_archive arch = { ARCH_ROMFS, { PATH_EMPTY, 1, (u8*)"" }, 0, 0 };
-		FS_path path = { PATH_BINARY, sizeof(zeros), zeros };
+		FS_Archive arch = { ARCHIVE_ROMFS, { PATH_EMPTY, 1, (u8*)"" }, 0 };
+		FS_Path path = { PATH_BINARY, sizeof(zeros), zeros };
 
-		Result rc = FSUSER_OpenFileDirectly(&romFS_file, arch, path, FS_OPEN_READ, FS_ATTRIBUTE_NONE);
+		Result rc = FSUSER_OpenFileDirectly(&romFS_file, arch, path, FS_OPEN_READ, 0);
 		if (R_FAILED(rc)) return rc;
 	}
 
--- a/libctru/include/3ds.h
+++ b/libctru/include/3ds.h
@@ -49,6 +49,7 @@ extern "C" {
 #include <3ds/services/mvd.h>
 #include <3ds/services/news.h>
 #include <3ds/services/qtm.h>
+#include <3ds/services/srvpm.h>
 #include <3ds/services/y2r.h>
 #include <3ds/services/hb.h>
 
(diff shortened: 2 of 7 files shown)


Lectem:
Could you take this opportunity to align enum members values ? (The = should be aligned too if possible)

Steveice10:
Like so?

fincs:
Just two complaints:

- Why were all these `Obsoleted` functions added? They are supposed to be obsolete.
- `ARCH_XYZ` was fine, I think it was unnecessary to rename it to `ARCHIVE_XYZ`.

mtheall:
PATH_CHAR: I would change the comment to say "ASCII text" instead of "8-bit text"
PATH_WCHAR: I would change the comment to say "UTF-16 text" instead of "16-bit text"

This is more explicit since there are several 8-bit and 16-bit character encodings. I might even go so far as to rename them to "PATH_ASCII" and "PATH_UTF16"

mtheall:
It would be useful to have in that header:
`extern FS_Attributes FS_ATTRIBUTES_NONE;`

and then have some file that defines it:
`FS_Attributes FS_ATTRIBUTES_NONE = { 0, 0, 0, 0 };`

Steveice10:
@fincs:
* They were added for completion. I can remove them if you wish.
* The attributes are actually a struct. A bitmask would be a misrepresentation.
* For clarity. Plus, arch is normally used as a short form of architecture. I can revert this if you'd like.

@mtheall Made the suggested changes.

mtheall:
Hmmm... Now I wonder is it okay to leave FS_Attributes as that struct or would it be better for it to become a 32-bit enum bitmask.

In any case, I probably wouldn't put FS_ATTRIBUTES_NONE in the sdmc_dev.c file (maybe it can go in its own file) because if your code uses it, that will pull in the whole sdmc_dev code whether you use it or not. On the other hand, I don't know how much code *doesn't* use sdmc_dev. This could be avoided if it becomes the 32-bit enum bitmask. I'll leave it to others to weigh in.

Steveice10:
@mtheall It's defined within fs.c. sdmc_dev.c only makes use of it.

mtheall:
Oh yeah you're right. I misread the line in sdmc_dev.c and the changes for fs.c were hidden since they changeset was so large, so I didn't see it there.

"it should be a struct as that's what it actually is" who decides that? The bitmask and the struct will have the same physical layout in memory, so unless there is some canonical source that says so, I think it can go either way.

    enum {
        FS_ATTRIBUTE_DIRECTORY = BIT(24),
        FS_ATTRIBUTE_HIDDEN    = BIT(16),
        FS_ATTRIBUTE_ARCHIVE   = BIT(8),
        FS_ATTRIBUTE_READ_ONLY = BIT(0),
    };
    #define FS_ATTRIBUTE_NONE 0

Really the part that gets tricky is: does any non-zero value in the 8-bit field mean it's turned on? Is there ever any case in real life where we could get an 8-bit value that is neither 0 nor 1 in such a field? If so, I would argue that the struct is more accurate. Either way, the implementations for FSUSER etc that use these attributes can transform the result of the syscall to give us these guarantees.

Steveice10:
@mtheall In disassembled game code it's treated as a struct, so I guess Nintendo decides that. :P

fincs:
> They were added for completion, and they do still work. I can remove them if you wish.

Yes, please.

As for the attributes thing: I side with mtheall. Regardless of what Nintendo does, I think it is just plain inconvenient for it to be a struct. A bitmask is easier to use and pass, and does not have the problems mtheall pointed out. I heavily doubt any of those fields get a non-zero-non-one value, but it can still be supported with a bitmask.

Steveice10:
Changes made.

libctru/source/romfs_dev.c
```diff
@@ -137,8 +137,8 @@ Result romfsInit(void)
 		if (units == (size_t)-1) return 3;
 		__utf16path[units] = 0;
 
-		FS_archive arch = { ARCH_SDMC, { PATH_EMPTY, 1, (u8*)"" }, 0, 0 };
-		FS_path path = { PATH_WCHAR, (units+1)*2, (u8*)__utf16path };
+		FS_Archive arch = { ARCHIVE_ROMFS, { PATH_EMPTY, 1, (u8*)"" }, 0, 0 };
```
fincs: Whoops, this got copypasted (should have been ARCHIVE_SDMC).

Steveice10: Fixed.

Steveice10:
In relation to the discussion on IRC about pointer casting, what if something like this were created:
```c
#define IPC_WRITE8(buf, i, v)  (buf)[i] = (v) & 0xFF
#define IPC_WRITE16(buf, i, v) (buf)[i] = (v) & 0xFFFF
#define IPC_WRITE32(buf, i, v) (buf)[i] = (v)
#define IPC_WRITE64(buf, i, v) (buf)[i] = (u32) (v); (buf)[i + 1] = (u32) ((v) >> 32)
#define IPC_WRITE(buf, i, ptr, size) memcpy(&((buf)[i]), (ptr), (size))

#define IPC_READ8(buf, i, ptr)  *(ptr) = (u8) ((buf)[i] & 0xFF)
#define IPC_READ16(buf, i, ptr) *(ptr) = (u16) ((buf)[i] & 0xFFFF)
#define IPC_READ32(buf, i, ptr) *(ptr) = (buf)[i]
#define IPC_READ64(buf, i, ptr) *(ptr) = (buf)[i] | ((buf)[i] << 32)
#define IPC_READ(buf, i, ptr, size) memcpy((ptr), &((buf)[i]), (size))
```

This would prevent https://github.com/smealum/ctrulib/issues/217 and provide clean standard macros for writing data larger than 32-bits.

Steveice10:
Either way, for now the size and pointer casting issues should be taken care of. Let me know if there are any further issues.

PR #219, opened 2015-11-15
Reference count srv, cleanup launcher code.

Steveice10:
Adds reference counting to SRV and consolidates launcher information access into a single set of functions, rather than a bunch of extern definitions everywhere.

### DIFF
--- a/libctru/include/3ds.h
+++ b/libctru/include/3ds.h
@@ -18,6 +18,7 @@ extern "C" {
 #include <3ds/synchronization.h>
 #include <3ds/gfx.h>
 #include <3ds/console.h>
+#include <3ds/env.h>
 #include <3ds/util/utf.h>
 
 #include <3ds/allocator/linear.h>
--- a/libctru/include/3ds/env.h
+++ b/libctru/include/3ds/env.h
@@ -1,0 +1,72 @@
+/**
+ * @file env.h
+ * @brief Homebrew environment information.
+ */
+#pragma once
+
+/// System run-flags.
+enum {
+	RUNFLAG_APTWORKAROUND = BIT(0), ///< Use APT workaround.
+	RUNFLAG_APTREINIT     = BIT(1), ///< Reinitialize APT.
+};
+
+/**
+ * @brief Gets whether the application was launched from a homebrew environment.
+ * @return Whether the application was launched from a homebrew environment.
+ */
+static inline bool envIsHomebrew(void) {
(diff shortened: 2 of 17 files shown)


fincs:
Just a few concerns:

* I wouldn't use the name `launcher` - hbmenu is actually not responsible for anything. It is the homebrew exploit that is responsible for implementing the protocol.
* The header file should be public.
* Why was `__linear_heap` renamed to `__ctru_linear_heap`?
* `launcherGetAptAppId` and everything below that should be static inline. The global variables they access should be defined inside it to maintain encapsulation like so:

```c
static inline u32 launcherGetAptAppId(void) {
	extern u32 __apt_appid;
	return __apt_appid;
}
```

Steveice10:
* What name would you recommend? I couldn't think of what else to use.
* I went ahead and documented it, but are you sure it should be public? It's pretty much all internal information for the *hax environment.
* To separate the fixed *hax parameters from ctrulib's heap management.
* Fixed.
* Fixed.

fincs:
> What name would you recommend? I couldn't think of what else to use.

I was thinking of something like `hbEnv` (for homebrew environment), or `hbLdr` (for homebrew loader). Note that the name `loader` is already taken by Nintendo :3. Either way we can brainstorm this on IRC.

> It's pretty much all internal information for the *hax environment.

It's not internal, it's part of the homebrew loading protocol. I expect functions like launcherIsPresent and launcherGetHandle to be useful to the general public (e.g. for using `hb:mem0`).

> To separate the fixed *hax parameters from ctrulib's heap management, for the size variables.

Oh I see. If the user overrides the default memory management code, modules such as the linearmem manager still use the old heap size.

Steveice10:
Building on your name suggestion, just "env" could work. What do you think?

Moved the header to the include directory so it's public.

fincs:
`env` is a better name (`launcherIsPresent` should become `envIsHomebrew` or something similar). BTW, 3ds.h needs to be updated; and `launcherDestroyHandles` should probably not be exposed in the header.

> (by the way, what IS `hb:mem0` anyway?)

It is an extra memory block stolen from SYSTEM that hax 2.5 passes to homebrew, however we still have not found a good way to integrate it seamlessly into existing infrastructure due to differences in behaviour when using memory SVCs; so for the moment it is unused. However, applications that want the memory anyway should be able to use it, even if manually.

Steveice10:
Changes made.

PR #220, opened 2015-11-15
Deprecate HB service.

Steveice10:
The HB service is only available from the outdated Ninjhax 1.1, and should therefore be deprecated to discourage use.

### DIFF
--- a/libctru/include/3ds/services/hb.h
+++ b/libctru/include/3ds/services/hb.h
@@ -10,21 +10,31 @@
 
 #include <3ds/types.h>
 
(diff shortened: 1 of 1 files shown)


fincs:
I think HB shouldn't be deprecated because hax 2.x doesn't support consoles below 9.0, and 1.1b is still useful for applications that require dynarec (even if that limits the highest system version to 9.2).

Steveice10:
By now, most dynarec homebrew applications use kernel access instead of HB since it's more flexible, but I can see where you're coming from.

PR #223, opened 2015-11-25
GPU: Separate output buffers and viewport

DiscostewSM:
Also allow setting output buffer formats


fincs:
The GPU code/API in the master branch is obsolete in favour of external GPU wrapper libraries.

PR #242, opened 2016-01-14
More soc

Lectem:
The following PR adds support for getaddrinfo and getnameinfo.
Note that the PF_INET6 value was checked thanks to genameinfo, which allows IPv6 DNS reverse lookup, but sockaddr_in6 wasn't added since it would lead people to think it works with other functions such as 'socket'.
Some EAI_* error codes are missing but I couldn't produce any other value.

Thanks to @mtheall for his help !

### DIFF
--- a/libctru/include/netdb.h
+++ b/libctru/include/netdb.h
@@ -17,6 +17,16 @@ struct hostent {
 	char	*h_addr;
 };
 
+
+#define AI_PASSIVE     0x01
+#define AI_CANONNAME   0x02
+#define AI_NUMERICHOST 0x04
+#define AI_NUMERICSERV 0x00 /* probably 0x08 but services names are never resolved */
+
+// doesn't apply to 3ds
+#define AI_ADDRCONFIG  0x00
+
+
 #define NI_MAXHOST     1025
 #define NI_MAXSERV       32
 
@@ -31,6 +41,16 @@ struct hostent {
 #define EAI_NONAME   (-305)
 #define EAI_SOCKTYPE (-307)
 
+struct addrinfo {
+	int             ai_flags;
+	int             ai_family;
+	int             ai_socktype;
+	int             ai_protocol;
+	socklen_t       ai_addrlen;
+	char            *ai_canonname;
+	struct sockaddr *ai_addr;
+	struct addrinfo *ai_next;
+};
 
 #ifdef __cplusplus
 extern "C" {
@@ -45,6 +65,12 @@ extern "C" {
 	int getnameinfo(const struct sockaddr *sa, socklen_t salen,
 		char *host, socklen_t hostlen,
 		char *serv, socklen_t servlen, int flags);
+
+	int getaddrinfo(const char *node, const char *service,
+		const struct addrinfo *hints,
+		struct addrinfo **res);
+
+	void freeaddrinfo(struct addrinfo *ai);
 
 #ifdef __cplusplus
 }
--- a/libctru/include/netdb.h
+++ b/libctru/include/netdb.h
@@ -17,6 +17,21 @@ struct hostent {
 	char	*h_addr;
 };
 
+#define NI_MAXHOST     1025
+#define NI_MAXSERV       32
+
+#define NI_NOFQDN      0x01
+#define NI_NUMERICHOST 0x02
+#define NI_NAMEREQD    0x04
+#define NI_NUMERICSERV 0x00 /* probably 0x08 but services names are never resolved */
+#define NI_DGRAM       0x00 /* probably 0x10 but services names are never resolved */
+
+#define EAI_FAMILY   (-303)
+#define EAI_MEMORY   (-304)
+#define EAI_NONAME   (-305)
+#define EAI_SOCKTYPE (-307)
+
+
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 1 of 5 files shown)


libctru/include/netdb.h
```diff
@@ -8,6 +8,8 @@
 #define NO_RECOVERY	3
 #define TRY_AGAIN	4
 
+
+
```
mtheall: Too much whitespace

libctru/include/netdb.h
```diff
@@ -17,6 +19,41 @@ struct hostent {
 	char	*h_addr;
 };
 
+
+#define AI_PASSIVE     0x01
+#define AI_CANONNAME   0x02
+#define AI_NUMERICHOST 0x04
+#define AI_NUMERICSERV 0x08
```
mtheall: Can we know the value of AI_NUMERICSERV?

libctru/include/netdb.h
```diff
@@ -17,6 +19,41 @@ struct hostent {
...
+#define NI_NOFQDN      0x01
+#define NI_NUMERICHOST 0x02
+#define NI_NAMEREQD    0x04
+#define NI_NUMERICSERV 0x00 /* probably 0x08 but services names are never resolved */
+#define NI_DGRAM       0x00 /* probably 0x10 but services names are never resolved */
+
+#define EAI_FAMILY   -303
+#define EAI_MEMORY   -304
+#define EAI_NONAME   -305
+#define EAI_SOCKTYPE -307
```
mtheall: I would put these values in parentheses, e.g. (-303)

libctru/include/netdb.h
```diff
@@ -17,6 +19,41 @@ struct hostent {
...
+struct addrinfo {
+  int ai_flags;
+  int ai_family;
+  int ai_socktype;
+  int ai_protocol;
+  socklen_t ai_addrlen;
+  char *ai_canonname;
+  struct sockaddr *ai_addr;
+  struct addrinfo *ai_next;
+};
```
mtheall: Please align the member names.

libctru/source/services/soc/soc_gai_strerror.c
```diff
@@ -0,0 +1,20 @@
+#include "soc_common.h"
```
mtheall: I don't think you need to include "soc_common.h"

libctru/source/services/soc/soc_gai_strerror.c
```diff
@@ -0,0 +1,20 @@
...
+#include <netdb.h>
+#include <stdio.h>
```
mtheall: I don't think you need to include <stdio.h>

libctru/source/services/soc/soc_getaddrinfo.c
```diff
@@ -0,0 +1,161 @@
...
+	struct addrinfo *next_ai = ai;
+	while(ai != NULL)
+	{
+		next_ai = ai->ai_next;
+		free(ai);
+		ai = next_ai;
+	}
+}
+
+struct addrinfo * buffer2addrinfo(addrinfo_3ds_t * entry)
```
mtheall: This should be a static function

libctru/source/services/soc/soc_getaddrinfo.c
```diff
@@ -0,0 +1,161 @@
...
+{
+	int ai_canonname_len = strnlen(entry->ai_canonname, sizeof(entry->ai_canonname));
+	struct addrinfo *ai;
+	size_t          len = sizeof(*ai)
+	                    + sizeof(struct sockaddr_storage)
+	                    + ai_canonname_len
+	                    + 1;
+
+	ai = (struct addrinfo*)calloc(len,1);
```
mtheall: This should be calloc(1, len)

libctru/source/services/soc/soc_getaddrinfo.c
```diff
@@ -0,0 +1,161 @@
...
+		ai->ai_addrlen  = entry->ai_addrlen;
+
+		memcpy(ai->ai_canonname, entry->ai_canonname, ai_canonname_len);
+		memcpy(ai->ai_addr, &entry->ai_addr, ai->ai_addrlen);
+	}
+	return ai;
+}
+
+
+int getaddrinfo_detail(const char *node, const char *service, const struct addrinfo *hints, struct addrinfo **res,addrinfo_3ds_t *info,s32 info_count, s32 * count)
```
mtheall: Put a space after each comma in the argument list

mtheall: This function should be static

libctru/source/services/soc/soc_getaddrinfo.c
```diff
@@ -0,0 +1,161 @@
...
+		*res = NULL;
+	else if(*count <= info_count)
+	{
+		struct addrinfo **ptr = res;
+		for(i = 0; i < *count; ++i)
+		{
+			*ptr = buffer2addrinfo(&info[i]);
+			if(*ptr == NULL)
+			{
+				freeaddrinfo(*res);
```
mtheall: I would also set *res = NULL here

libctru/include/netdb.h
Lectem: Not really, it was just a guess, I'll change it to 0x00

Lectem:
I addressed all the comments

PR #243, opened 2016-01-16
fixed usleep

Lectem:
the value of useconds * 1000 is u32, while we want u64


Lectem:
done in 557fbedb9b7a837f5a39f3eeb3c7c173b24beddf

PR #252, opened 2016-01-22
Added some missing news:s functions.

Rinnegatamante:
Added some missing news:s functions.

### DIFF
--- a/libctru/include/3ds/services/news.h
+++ b/libctru/include/3ds/services/news.h
@@ -44,9 +44,9 @@ Result NEWS_GetTotalNotifications(u32* num);
 /**
  * @brief Sets a custom header for a specific notification.
  * @param news_id Identification number of the notification.
- * @param header Notification header to set.
+ * @param header Pointer to the notification header to set.
  */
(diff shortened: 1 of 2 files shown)


libctru/include/3ds/services/news.h
```diff
@@ -21,3 +34,38 @@ void newsExit(void);
...
+ * @param num Pointer where total number will be saved.
+ */
+Result NEWS_GetTotalNotifications(u32* num);
+
+/**
+ * @brief Sets a custom header for a specific notification.
+ * @param news_id Identification number of the notification.
+ * @param header Notification header to set.
+ */
+Result NEWS_SetNotificationHeader(u32 news_id, NotificationHeader header);
```
fincs: Should take `const NotificationHeader*` instead of a structure by value.

Steveice10:
While your at it, I would recommend updating the NotificationHeader structure to the latest 3dbrew documentation, if it wouldn't be much trouble.

http://3dbrew.org/wiki/NEWSS:AddNotification#Header_structure

Rinnegatamante:
Going to modify header and re-pull it later.

PR #253, opened 2016-01-23
Added svc calls for svcCreateCodeSet and svcCreateProcess

yifanlu:
Added these two syscalls

### DIFF
--- a/libctru/include/3ds/svc.h
+++ b/libctru/include/3ds/svc.h
@@ -241,6 +241,30 @@ typedef struct {
 
 ///@}
 
+///@name Processes
+///@{
+
+/// Information on address space for process. All sizes are in pages (0x1000 bytes)
+typedef struct {
+	u8 name[8];           ///< ASCII name of codeset
+	u16 unk1;
+	u16 unk2;
+	u32 unk3;
+	u32 text_addr;        ///< .text start address
+	u32 text_size;        ///< .text number of pages
+	u32 ro_addr;          ///< .rodata start address
+	u32 ro_size;          ///< .rodata number of pages
+	u32 rw_addr;          ///< .data, .bss start address
+	u32 rw_size;          ///< .data number of pages
+	u32 text_size_total;  ///< total pages for .text (aligned)
+	u32 ro_size_total;    ///< total pages for .rodata (aligned)
+	u32 rw_size_total;    ///< total pages for .data, .bss (aligned)
+	u32 unk4;
+	u64 program_id;       ///< Program ID
+} CodeSetInfo;
+
+///@}
+
 /**
  * @brief Gets the thread local storage buffer.
  * @return The thread local storage bufger.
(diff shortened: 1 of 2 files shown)


libctru/include/3ds/svc.h
```diff
@@ -241,6 +241,30 @@ typedef struct {
 
 ///@}
 
+///@name Processes
+///@{
+
+/// Information on address space for process. All sizes are in pages (0x1000 bytes)
+typedef struct {
+	u8 name[8];		///< ASCII name of codeset
```
fincs: Please do not use tabs to align comments; use spaces instead.

libctru/include/3ds/svc.h
```diff
@@ -241,6 +241,30 @@ typedef struct {
...
+	u16 unk1;
+	u16 unk2;
+	u32 unk3;
+	u32 text_addr;				///< .text start address
```
fincs: This should be extracted to a separate struct:

```c
typedef struct
{
	u32 addr;
	u32 size;
} CodeSetSecInfo;
```
and it would be used like this:
```
CodeSetSecInfo code;
CodeSetSecInfo rodata;
CodeSetSecInfo data;
```

libctru/include/3ds/svc.h
```diff
@@ -241,6 +241,30 @@ typedef struct {
...
+	u32 text_size;				///< .text number of pages
+	u32 ro_addr;					///< .rodata start address
+	u32 ro_size;					///< .rodata number of pages
+	u32 rw_addr;					///< .data, .bss start address
+	u32 rw_size;					///< .data number of pages
+	u32 text_size_total;	///< total pages for .text (aligned)
+	u32 ro_size_total;		///< total pages for .rodata (aligned)
+	u32 rw_size_total;		///< total pages for .data, .bss (aligned)
+	u32 unk4;
+	u8 program_id[8];			///< Program ID
```
fincs: This should be `u64 title_id` instead of an array.

libctru/include/3ds/svc.h
```diff
@@ -487,6 +511,25 @@ Result svcCreatePort(Handle* portServer, Handle* portClient, const char* name, s
...
+
+/**
+ * @brief Sets up virtual address space for a new process
+ * @param[out] out Pointer to output the code set handle to.
+ * @param info Description for setting up the addresses
+ * @param code_ptr Pointer to .text in shared memory
+ * @param ro_ptr Pointer to .rodata in shared memory
+ * @param data_ptr Pointer to .data in shared memory
+ */
+Result svcCreateCodeSet(Handle* out, CodeSetInfo *info, void* code_ptr, void* ro_ptr, void* data_ptr);
```
fincs: CodeSetInfo pointer should be const.

libctru/include/3ds/svc.h
yifanlu: I was going to do that, but it adds confusion to the _total members. My code is like this
```
info.ro_addr = ...;
info.ro_size = ...;
info.ro_size_total = ...;
```
Since the proposed struct doesn't seem to be used anywhere else, I think it's cleaner to leave it as is.

PR #255, opened 2016-01-23
Added ld flags for link time optimizations and separating data sections

yifanlu:
For anyone who has size constraints on their output binaries, adding these made my executable 28KB -> 26KB. Not a lot, but it helps. -fdata-sections currently does not do anything, but it doesn't hurt to have the option there.

### DIFF
--- a/libctru/Makefile
+++ b/libctru/Makefile
@@ -42,9 +42,9 @@ INCLUDES	:=	include
 #---------------------------------------------------------------------------------
 ARCH	:=	-march=armv6k -mtune=mpcore -mfloat-abi=hard -mtp=soft
 
-CFLAGS	:=	-g -Wall -Werror -O2 -mword-relocations \
-			-ffunction-sections -fno-strict-aliasing \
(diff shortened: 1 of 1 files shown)


fincs:
`-Os` should *not* be the default since it makes GCC emit slower code. Additionally, `-fdata-sections` was deliberately left out because there were problems with it. Finally, while enabling LTO is interesting, it looks like it is not yet ready in GCC (I recall having the compiler crash when trying to build portal3DS with LTO); and what's even more, `-flto` alone produces pure LTO objects that cannot be used in projects that do not use LTO.

All in all, when considering the only improvement is a meager 2KB, I do not see how any of these flags have more benefit than potential cost.

profi200:
Never had -flto crashing with anything here. The only problem was that non-lto aware tools were used before the latest devkitARM iirc.

yifanlu:
That's interesting that `-fdata-sections` had issues. Turns out, I can get the bulk of the code to go away with custom init stuff, so this is no longer needed.

PR #256, opened 2016-01-25
added SOCU_GetNetworkOpt

Lectem:
This commit adds SOCU_GetNetworkOpt and some of the structs used to read the network configuration.


mtheall:
Can you suffix the types with _t?

Lectem:
I didn't add those since most of the ctrulib code base doesn't have suffixes for structs

PR #258, opened 2016-01-31
httpcOpenContext() can do more than just GET

ksanislo:
I've added an additional parameter to httpcOpenContext() that allows the creation of requests for GET POST HEAD PUT and DELETE. There's a possibility that there are still additional methods available, but things were a bit strange on higher numbers. 0x6 seemed to be another POST, and 0x7 retrieved a file without seeming to make an actual web request, cache return perhaps? Values above 0x7 all seemed to fail.

### DIFF
--- a/libctru/include/3ds/ndsp/channel.h
+++ b/libctru/include/3ds/ndsp/channel.h
@@ -81,6 +81,21 @@ u32  ndspChnGetSamplePos(int id);
  * @return The sequence ID of the wave buffer.
  */
 u16  ndspChnGetWaveBufSeq(int id);
+
+/**
+ * @brief Checks whether a channel is currently paused.
+ * @param id ID of the channel (0..23).
+ * @return Whether the channel is currently paused.
+ */
+bool ndspChnIsPaused(int id);
+
+/**
+ * @brief Sets the pause status of a channel.
+ * @param id ID of the channel (0..23).
+ * @param paused Whether the channel is to be paused (true) or unpaused (false).
+ */
+void ndspChnSetPaused(int id, bool paused);
+
 ///@}
 
 ///@name Configuration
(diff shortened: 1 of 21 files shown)


yellows8:
FWIW ctrulib has been rather outdated for HTTPC compared to wiki for a good while. https://3dbrew.org/wiki/HTTPC:CreateContext

ksanislo:
@yellows8 I'm working on implementing more of the functionality documented on the wiki in my fork, but I can't figure out which, if any, of those calls will return the size while using "Transfer-Encoding: chunked". GetDownloadSizeState only returns the full size from the header when non-chunked, and returns a 0 when using chunked encoding, but it must be getting processed off somewhere, since ReceiveData starts by returning the data without the chunk size value first.

I don't suppose there's a chance you have any insight into where the 3DS might be keeping this info?

yellows8:
I'm not seeing anything in this cmdlog for getting any size from HTTPC where chunked-encoding is used. No idea how Nintendo handles it.

ksanislo:
httpcDownloadData() no longer relies on knowing the size of what it needs to collect beforehand, and can handle downloads up to u32 size by calling repeatedly until it stops returning the status code for more data pending, and properly returns the byte counts downloaded on each call. This works identically no matter if you're using a normal HTTP download, or chunked encoding (so on-the-fly encoded gzip/deflate in streams can be used now!) I'd still like to implement the rest of PUT and POST at some point, but there's no reason this shouldn't be merged as-is, its far less broken than what was here to start with.

yellows8:
Can you git squash this? A PR to update this would be nice too: https://github.com/devkitPro/3ds-examples/tree/master/network/http

ksanislo:
@yellows8, I've done a rebase and push to hopefully squash my commits together, I'm not terribly familiar with that functionality though, so I may not have done it correctly. Let me know if there are any problems with that.

And I've already got a patched fork for the 3ds-examples http stuff that I'll make a PR on right now.

PR #259, opened 2016-02-03
Fix crash when getting or setting struct params in CAM:U

yuriks:


mtheall:
Line 633: you can't assign to a const reference

yuriks:
@mtheall Just pushed a redone version.

PR #260, opened 2016-02-04
Add basic NFC service support

javimadgit:


fincs:
Looks good. The only complaint is that service call wrappers should be prefixed with `NFC_` and not `nfc_`.

javimadgit:
Fixed prefix and parameters declaration.

PR #261, opened 2016-02-07
Clean up and update AM service functions.

Steveice10:
Cleans up AM functions, and both brings existing ones up to date with and adds new functions from 3dbrew documentation.

### DIFF
--- a/libctru/include/3ds/services/am.h
+++ b/libctru/include/3ds/services/am.h
@@ -4,9 +4,9 @@
  */
 #pragma once
 
-/**
- * @brief Contains basic information about a title.
- */
+#include <3ds/services/fs.h>
+
+/// Contains basic information about a title.
 typedef struct
 {
 	u64 titleID; ///< The title's ID.
@@ -14,6 +14,39 @@ typedef struct
 	u16 version; ///< The title's version.
 	u8 unk[6];   ///< Unknown title data.
 } AM_TitleEntry;
+
+/// Pending title status mask values.
+enum
+{
+	AM_STATUS_MASK_INSTALLING = BIT(0),           ///< Titles currently installing.
+	AM_STATUS_MASK_AWAITING_FINALIZATION = BIT(1) ///< Titles awaiting finalization.
+};
+
+/// Pending title status values.
+typedef enum
+{
+	AM_STATUS_ABORTED = 0x0002,              ///< Install aborted.
+	AM_STATUS_SAVED = 0x0003,                ///< Title saved, but not installed.
+	AM_STATUS_INSTALL_IN_PROGRESS = 0x0802,  ///< Install in progress.
+	AM_STATUS_AWAITING_FINALIZATION = 0x0803 ///< Awaiting finalization.
+} AM_InstallStatus;
+
+// Contains basic information about a pending title.
+typedef struct
+{
+    u64 titleId;   ///< Title ID
+    u16 version;   ///< Version
+    u16 status;    ///< @ref AM_InstallStatus
+    u32 titleType; ///< Title Type
+    u8 unk[0x8];   ///< Unknown
+} AM_PendingTitleEntry;
+
+/// Pending title deletion flags.
+enum
+{
+	AM_DELETE_PENDING_NON_SYSTEM = BIT(0), ///< Non-system titles.
+	AM_DELETE_PENDING_SYSTEM = BIT(1)      ///< System titles.
+};
 
 /// Initializes AM.
 Result amInit(void);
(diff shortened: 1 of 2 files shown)


libctru/include/3ds/services/am.h
```diff
@@ -25,41 +58,84 @@ void amExit(void);
...
+ * @param[out] ticketIds Pointer to output the retrieved ticket IDs to.
+ * @param[out] ticketsRead Pointer to output the number of read tickets to.
+ * @param ticketCount Number of tickets to read.
+ * @param skip Number of tickets to skip.
+ */
+Result AM_GetTicketList(u64 *ticketIds, u32 *ticketsRead, u32 ticketCount, u32 skip);
+
+/**
+ * @brief Gets the number of pending titles on this system.
+ * @param[out] total Pointer to output the total to.
```
profi200: Should not this be "count"?

libctru/include/3ds/services/am.h
```diff
@@ -25,41 +58,84 @@ void amExit(void);
...
 
 /**
- * @brief Gets the number of titles for a given mediatype.
- * @param mediatype Mediatype to get titles from.
+ * @brief Gets the number of titles for a given media type.
+ * @param mediatype Media type to get titles from.
  * @param count Pointer to write the title count to.
  */
-Result AM_GetTitleCount(u8 mediatype, u32 *count);
+Result AM_GetTitleCount(u32 *count, FS_MediaType mediatype);
```
Lectem: any reason for switching the arguments order ?

Steveice10: Seemed like a good idea to standardize the functions on outputs followed by inputs. I can revert these kinds of changes if you'd like.

Lectem: well, the thing is that it's breaking the API while there are other AM_functions that take the mediatype as first argument (AM_DeletePendingTitles,AM_DeletePendingTitle ...)
While this is only personal preference, I usually like to have the mode/type/enum first so that you know what value you will be getting from the following parameters.

Steveice10: Taken care of.

libctru/include/3ds/services/am.h
Steveice10: Fixed.

fincs:
We don't care about breaking the API. Go ahead and place output arguments before input arguments.

Lectem:
My point was that it was not consistent with the rest of the file, and therefore didn't justify the API break (see https://github.com/Steveice10/ctrulib/commit/bac29656a57e954f47cefc124408609d142ca86a#diff-e929320527f19526d7fd64e2c95a107aR98 and https://github.com/Steveice10/ctrulib/commit/bac29656a57e954f47cefc124408609d142ca86a#diff-e929320527f19526d7fd64e2c95a107aR125 for example)

Steveice10:
@Lectem Those actually were changed, but I changed them again after your request. Also, it actually still is how I mentioned. Notice how in AM_GetTicketList, "ticketsRead" is first. I actually changed "ticketIds", etc. to not be first as they're technically parameters, not cmdbuf outputs.

Lectem:
I don't really understand why ticketIds shouldn't be considered as output from an API point of view since the documentation says "Buffer to output the retrieved ticket IDs to."

Steveice10:
So do you want outputs moved in front or not? Make up your mind. :P

Steveice10:
Any further edits needed to get this pulled?

mariogamer2:
This seem to brick some build (like sysUpdater) (When I switched to the normal ctrulib,it worked...)

profi200:
What did you expect if the API got changed?

urherenow:
@mariogamer2 sysupdater would have to be updated. I imagine my reaction just now was about the same as @profi200 's

libctru/include/3ds/services/am.h
```diff
@@ -25,96 +58,175 @@ void amExit(void);
...
- * @param mediatype Mediatype of the title.
+ * @param mediatype Media type of the title.
+ * @param titleID ID of the title.
+ * @param[out] productCode Pointer to output the product code to. (length = 16)
+ */
+Result AM_GetTitleProductCode(FS_MediaType mediatype, u64 titleId, char *productCode);
+
+/**
+ * @brief Gets the ext data ID of a title.
+ * @param[out] productCode Pointer to output the ext data ID to.
```
profi200: Copy paste fail.

urherenow:
So... What is wrong with this that it isn't pushed yet?

PR #265, opened 2016-02-16
add a note for svcExitThread and the thread API

Lectem:
Calling svcExitThread will make it so that threadCreate detached parameter won't work.

### DIFF
--- a/libctru/include/3ds/thread.h
+++ b/libctru/include/3ds/thread.h
@@ -31,6 +31,9 @@ typedef struct Thread_tag* Thread;
  * - Processor #2 is New3DS exclusive. Normal applications can create threads on this core if the exheader kernel flags bitmask has 0x2000 set.
  * - Processor #3 is New3DS exclusive. Normal applications cannot create threads on this core.
  * - Processes in the BASE memory region can always create threads on processors #2 and #3.
+ *
+ * @note Default exit code of a thread is 0.
+ * @warning @ref svcExitThread should never be called from the thread, use @ref threadExit instead.
  */
 Thread threadCreate(ThreadFunc entrypoint, void* arg, size_t stack_size, int prio, int affinity, bool detached);
 


libctru/include/3ds/thread.h
```diff
@@ -31,6 +31,9 @@ typedef struct Thread_tag* Thread;
  * - Processor #2 is New3DS exclusive. Normal applications can create threads on this core if the exheader kernel flags bitmask has 0x2000 set.
  * - Processor #3 is New3DS exclusive. Normal applications cannot create threads on this core.
  * - Processes in the BASE memory region can always create threads on processors #2 and #3.
+ * 
+ * @note For the detached parameter to work, you need to either return from your entrypoint or call threadExit. 
+ *       svcExitThread should not be used, or you will have to call threadFree yourself.
```
fincs: This note can be summed up with: `svcExitThread should never be used inside the thread; use threadExit instead.`. No need to talk about an implementation detail not working if you don't do the right thing.

Also you can't free yourself using `threadFree`.

Lectem: Should I leave the part about using return ?
Obviously I didn't mean calling threadFree from the thread itself, but it could indeed be misleading, I will remove it anyway.

fincs: I meant the whole note. The detached parameter will only fail to work if you bypass internal management by using svcExitThread.

libctru/include/3ds/thread.h
```diff
@@ -23,7 +23,7 @@ typedef struct Thread_tag* Thread;
  * @param affinity The ID of the processor the thread should be ran on. Processor IDs are labeled starting from 0.
  *                 On Old3DS it must be <2, and on New3DS it must be <4.
  *                 Pass -1 to execute the thread on all CPUs and -2 to execute the thread on the default CPU (read from the Exheader).
- * @param detached When set to true, the thread is automatically freed when it finishes.
+ * @param detached When set to true, the thread is automatically freed when it finishes. Exit code will be 0.
```
Lectem: This comment should actually be another note, since it is always true. I'd like to change it to `@note Exit code will be 0 if threadExit isn't called.`, what do you think ?

fincs: `The default exit code of a thread is 0` explains what's going on better IMO.

libctru/include/3ds/thread.h
Lectem: sure, one last thing, should I actually make it a `@warn` instead of `@note` ?

fincs: Makes sense.

Lectem:
Should be ok now

fincs:
No mention of threadExit?

Lectem:
right, updated

PR #266, opened 2016-02-24
Update hid.h comments

ObsidianX:
Nit-picky update to accurately describe the `hidKeys*()` functions, and how to extract individual bits from them.


profi200:
Don't you think and'ing the value with the wanted button macros should be common knowledge? Not saying your commit is not useful.

ObsidianX:
Oh sure, but the current comment is incorrect by suggesting OR instead of AND.

PR #267, opened 2016-03-02
Adds function addPostDataRaw

Bownairo:
Implements the function addPostDataRaw for putting raw data into to body of a post request.

### DIFF
--- a/libctru/include/3ds/services/httpc.h
+++ b/libctru/include/3ds/services/httpc.h
@@ -68,6 +68,14 @@ Result httpcAddRequestHeaderField(httpcContext *context, char* name, char* value
 Result httpcAddPostDataAscii(httpcContext *context, char* name, char* value);
 
 /**
+ * @brief Adds a POST body to a HTTP context.
+ * @param context Context to use.
+ * @param data The data to be passed as raw into the body of the post request.
+ * @param len Length of data passed by data param.
+ */
+Result httpcAddPostDataRaw(httpcContext *context, u32* data, u32 len);
+
+/**
  * @brief Begins a HTTP request.
  * @param context Context to use.
  */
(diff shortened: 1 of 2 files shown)


Lectem:
Please edit your commit so that there are no linespace changes (use git commit --amend and git push -f)

yellows8:
This is missing a httpc* wrapper func which the other context-specific code use as well.

Bownairo:
Did you ever make a similar change when implementing addPostDataAscii?

yellows8:
"Did you ever make a similar change when implementing addPostDataAscii?" https://github.com/smealum/ctrulib/blob/b0c5a3c87d570508e96fb1e6718b8e55c678a720/libctru/source/services/httpc.c#L124

Bownairo:
I'm pretty positive I already have that implemented, are you sure I'm missing something?

yellows8:
Those comments in the .h are still wrong, please base it on 3dbrew. You should not be handling "Raw" data as a string either.

Bownairo:
I redid the params to pass the data as a u32 pointer directly instead of a string. Where are my comments still wrong?

profi200:
For unknown RAW data void* is good.

Bownairo:
From the looks of it, everything gets passed to cmdbuff as a u32 anyway, but it's not a hard fix to use a void pointer instead, if that's a more appropriate solution.

yellows8:
"Where are my comments still wrong?" NVM, thought this was another command which was actually documented on 3dbrew. Please document it on 3dbrew.

yellows8:
These "Lenght" typos need fixed.

Bownairo:
Typo has been fixed and wiki has been updated.

yellows8:
The .h comments don't really match what you added to the wiki, the former should be updated("form field").

Bownairo:
I changed "body" to "form field." What can I change next?

yellows8:
I'm not sure what "form" is supposed to mean here when there's no name-buffer used with that command.

Bownairo:
I changed everything back to POST body.

PR #282, opened 2016-04-29
Separate information required to open an archive from the actual opened handle.

Steveice10:
This separates the information required to open an archive from the actual opened handle, making it easier to pass around an opened archive in user code without having to worry about keeping path data, etc. around.

Because it would've been difficult to work this change around, the "sdmc" parameter to fsUseSession was also removed. I could not find any real use for it anyway, as you could easily just switch back to the main session with fsEndUseSession, if needed, to access the SD card.

### DIFF
--- a/libctru/source/services/fs.c
+++ b/libctru/source/services/fs.c
@@ -9,20 +9,30 @@
 #include <3ds/env.h>
 #include "../internal.h"
 
+#define FS_MAX_EXEMPT_ARCHIVE_HANDLES 16
+
 static Handle fsuHandle;
 static int fsuRefCount;
 
-static Handle fsSessionForArchive(FS_ArchiveID arch)
+static FS_Archive fsExemptArchives[FS_MAX_EXEMPT_ARCHIVE_HANDLES];
+
+static Handle fsSession(void)
 {
 	ThreadVars* tv = getThreadVars();
-	if (tv->fs_magic == FS_OVERRIDE_MAGIC && (arch != ARCHIVE_SDMC || tv->fs_sdmc))
+	if (tv->fs_magic == FS_OVERRIDE_MAGIC)
 		return tv->fs_session;
 	return fsuHandle;
 }
 
-static Handle fsSession(void)
+static Handle fsSessionForArchive(FS_Archive archive)
 {
-	return fsSessionForArchive(0);
+	for (int i = 0; i < FS_MAX_EXEMPT_ARCHIVE_HANDLES; i++)
+	{
+		if (fsExemptArchives[i] == archive)
+			return fsuHandle;
+	}
+
+	return fsSession();
 }
 
 Result fsInit(void)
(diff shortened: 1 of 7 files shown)


fincs:
This is an interesting change, and the API does feel cleaner with it. However...

> Because it would've been difficult to work this change around, the "sdmc" parameter to fsUseSession was also removed.

The sdmc parameter was added so that it's possible to use stdio/etc at the same time as fsUseSession. Applications like dumpers (uncart) will now need to use more complicated code. If we are removing the sdmc parameter, there should be a way to disable service handle redirection for specific archive handles (and the libctru sdmc code should register its handle in this list).

profi200:
"(uncart)" braindump

fincs:
Oops. Comment updated.

Steveice10:
Good idea. When I'm free this afternoon, I'll see if I can add in a way to exempt specific archive handles from sessions. The main issue I had with this before is how to reliably store it. However, considering FS has a limit on archive handles open simultaneously, maybe it would be okay to just have a fixed-length array of handles?

fincs:
Yeah. It can be a small array of 4 handles for example.

Steveice10:
Done.

libctru/source/services/fs.c
```diff
@@ -9,20 +9,30 @@
 #include <3ds/env.h>
 #include "../internal.h"
 
+#define FS_MAX_ARCHIVE_HANDLES 32
```
fincs: I don't think that big of a handle table is needed, especially considering linear lookup is used (I was earlier suggesting 4). Otherwise good job.

Steveice10: When I ran tests, I was able to create 27 archives before getting out of resource errors, and I assume the remaining 5 are in use by other processes and the SDMC driver.

fincs: The point is that it is extremely rare to want to bypass the session for so many archives.

profi200: But what IF? Then ctrulib needs to be changed again. I would make a compromise between size and usability like 15 handles.

fincs: I still don't see how anybody would want to bypass the session handle override for more than 4 archives. Note than in the previous system only a hardcoded bypass (for sdmc) was supported.

Steveice10: I'm perfectly fine with editing in a compromise between the two if you can give me a specific number.

urherenow:
Didn't he already though? Make it 16 and get 'er merged. If it needs to be increased later, then it's as simple as a fix could be.

Steveice10:
Done.

urherenow:
@fincs is this not yet good to merge?

neobrain:
Could we extend the documentation a bit? I have no idea what "Exempts an archive from using alternate FS session handles" is trying to say or under what circumstances I'm supposed to use it (I think that might be because the docs implicitly assume everybody knows what archives and their relation to sessions are) :/

Giving more background on archives and potentially giving crossreferences to other API calls would make this easier to grasp. Technically giving full background on archives may be out of scope for this PR, but when doing breaking changes we should at least aim for making it dead-obvious for software authors to adjust their projects accordingly ;)

By the way, minor nit: I don't think "unexempt" is a word.

einstein95:
Adding onto what @neobrain said, the comment "Exempts an archive from using alternate FS session handles" is difficult to understand, even though "unexempt" is a real word. Perhaps a more suitable term could be "free" instead of "exempt" and "doesn't free" instead of "unexempt"?

Steveice10:
@einstein95 "doesn't free" definitely isn't the right phrase for this; "exempt" means the archive will still use the main session handle instead of one specified by fsUseSession, and "unexempt" removes it from that list.

@neobrain I'm all for better docs, but this is not a doc PR.

neobrain:
@steveice10 Until a doc PR appears, I would prefer not to have this source-breaking change, then. It's tough enough for people to write programs given the current tools, so randomly breaking program sources without proper documentation isn't helping.

Steveice10:
@neobrain Added some clarification to the documentation of the main parts i changed.

PR #285, opened 2016-05-08
Allow toggling sleep mode.

Steveice10:
Makes it possible to disable sleep mode. When disabled, closing the lid will simply turn off the screen while still running code.


Cruel:
Is this at all related to the "Runnable on sleep" exheader flag? Or does that permit something else?

Steveice10:
@Cruel Don't know. This is basically just sending a rejection reply to sleep mode requests.

PR #289, opened 2016-05-31
Add pxi:dev service.

Steveice10:
Currently contains PXIDEV_SPIWriteRead and PXIDEV_SPIMultiWriteRead.

### DIFF
--- a/libctru/include/3ds/services/pxidev.h
+++ b/libctru/include/3ds/services/pxidev.h
@@ -1,0 +1,79 @@
+/**
+ * @file pxidev.h
+ * @brief Gamecard PXI service.
+ */
+#pragma once
+
+#include <3ds/services/fs.h>
+
+/// Card SPI wait operation type.
+typedef enum {
+	WAIT_NONE          = 0, ///< Do not wait.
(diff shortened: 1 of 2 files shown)


fincs:
In order to fit better with the rest of the codebase, PXIDEV_TransferOption and PXIDEV_WaitOperation shouldn't be bitfield structs, instead they should be normal integers and a static inline function for building the value should exist. Also, the two functions you added have a really high number of parameters. For convenience's sake you should probably put them into structs passed by pointer. I can already imagine a PXIDEV_Buffer struct. E.g. PXIDEV_SPIMultiWriteRead would be something this:

```c
Result PXIDEV_SPIMultiWriteRead(PXIDEV_Buffer* header, PXIDEV_Buffer writeBuffers[2], PXIDEV_Buffer readBuffers[2], PXIDEV_Buffer* footer);
```

This is just a suggestion, you can do something different if you think it's more convenient.

Steveice10:
Adjustments made, although I opted to keep the two write and read buffers separate to keep it in a natural order of what is written from/read to from left to right.

PR #301, opened 2016-07-26
Add httpcSetKeepAlive() to turn HTTP Keep-Alive on and off

ksanislo:
Being able to enable/disable Connection: Keep-Alive for HTTP is kind of a big deal.

### DIFF
--- a/libctru/include/3ds/services/httpc.h
+++ b/libctru/include/3ds/services/httpc.h
@@ -252,3 +252,10 @@ Result httpcCloseClientCertContext(u32 ClientCert_contexthandle);
  */
 Result httpcDownloadData(httpcContext *context, u8* buffer, u32 size, u32 *downloadedsize);
 
+/**
+ * @brief Sets Keep-Alive for the context.
+ * @param context Context to set flags on.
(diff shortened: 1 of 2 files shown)


yellows8:
What "option flags" exactly are should be mentioned in the comments.

ksanislo:
@yellows8 I'm not sure what all is available in here, as I have no idea how to determine how the API works other than trial and error. BIT(0) seems to be the bool for Keep-Alive on and off. You can set a HTTP header of 'Connection: keep-alive' and the 3DS will maintain the TCP connection for at least 5 seconds in case you make another context to the same server. Do you have any way to tell if there are any other features that can be controlled by this?

yellows8:
I've always assumed that entire field was just a bool(haven't checked sysmodule code for it iirc).

ksanislo:
I've documented the on/off(?) setting, however I would expect there to be some kind of a setting for a timeout and possibly max requests before requiring a new connection as well, so I doubt this is complete.

fincs:
This looks OK to me. Any objections before it is merged?

yellows8:
Looks fine.

PR #308, opened 2016-08-25
Added FSUSER_UpdateSha256Context().

profi200:
Tested and working with *hax.

### DIFF
--- a/libctru/source/services/fs.c
+++ b/libctru/source/services/fs.c
@@ -1201,6 +1201,27 @@ Result FSUSER_GetLegacySubBannerData(u32 bannerSize, FS_MediaType mediaType, u64
 	return cmdbuf[1];
 }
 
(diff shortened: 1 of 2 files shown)


libctru/source/services/fs.c
```diff
@@ -1201,6 +1201,27 @@ Result FSUSER_GetLegacySubBannerData(u32 bannerSize, FS_MediaType mediaType, u64
 	return cmdbuf[1];
 }
 
+Result FSUSER_UpdateSha256Context(const void* data, u32 inputSize, u8* hash)
+{
+	u32 *cmdbuf = getThreadCommandBuffer();
+
+	cmdbuf[0] = IPC_MakeHeader(0x84E, 13, 2); // 0x84E0342
+	cmdbuf[9] = inputSize;
```
Lectem: what about cmdbuf words from 1 to 8 ?

PR #314, opened 2016-09-17
Add support for nwm::EXT::ControlWirelessEnabled.

thedax:
https://www.3dbrew.org/wiki/NWMEXT:ControlWirelessEnabled

### DIFF
--- a/libctru/include/3ds/services/nwmext.h
+++ b/libctru/include/3ds/services/nwmext.h
@@ -1,0 +1,13 @@
+#pragma once
+
+// Initializes NWMEXT.
+Result nwmExtInit(void);
+
+// Exits NWMEXT.
+void nwmExtExit(void);
+
+/**
+ * @brief Turns wireless on or off.
+ * @param enableWifi True enables it, false disables it.
+ */
+Result NWMEXT_ControlWirelessEnabled(bool enableWifi);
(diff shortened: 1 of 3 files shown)


libctru/include/3ds/services/nwmext.h
```diff
@@ -0,0 +1,19 @@
...
+Result nwmExtInit(void);
+
+// Exits NWMEXT.
+void nwmExtExit(void);
+
+/**
+ * @brief Turns wireless on or off.
+ * @param command Command to turn wireless on or off.
+ */
+Result NWMEXT_ControlWirelessEnabled(u32 command);
```
fincs: Why isn't this just a bool?

thedax: As far as I can tell, just about all of the other services use Results, so why should this be any different? When I tested it a while back it was able to return error codes, not just a true or false boolean value.

fincs: I mean the input parameter, not the output parameter.

thedax: Oh, yeah. Sorry, I misunderstood. I can change that no problem.

thedax:
Rebased and squashed.

PR #318, opened 2016-10-19
Fixed buffer overflow after gfxSetScreenFormat

esoteric-programmer:


fox8091:
Why not fix the offending function (`__get_bytes_per_pixel`) instead of using a hardcoded value?

esoteric-programmer:
If one modifies `__get_bytes_per_pixel`, the only fix for the problem I found would be to always return 4.

However, this would break many other functions, like `aptInitCaptureInfo`.
I think there are two possible fixes without hard coding a value:
1. Compute the maximum over all values returned by `__get_bytes_per_pixel`
2. Fix `gfxSetScreenFormat` to reallocate buffer.
I don't like solution 1. I thought about solution 2 for a longer time, but I didn't really liked it, because
1. all buffers were initialized in gfx.c, regardless of double buffering or 3D being enabled or not and I felt that this would break the pattern,
2. `linearRealloc` is not implemented yet, so one has to use `linearFree` and `linearMalloc`.
However, I will modify the pull request if other people think that reallocating the buffers in `gfxSetScreenFormat` would be better.

fincs:
Does it really make sense to use RGBA as the pixel format for the LCDs? Keep in mind that the alpha information is completely ignored as far as we know.

esoteric-programmer:
I think, as long as ctrulib provides this option, it should not cause unpredictable memory access. One may remove RGBA8 from ctrulib and fix `gfxInit` with 3 as a constant instead (or `__get_bytes_per_pixel(GSP_BGR8_OES)`, or whatever).
However, note that the bug I'm addressing here does not only occur if one switches to RGBA8: One may initialize gfx with frame format `GSP_RGB565_OES` and switch to `GSP_BGR8_OES` later – no alpha channels. This will also trigger unpredictable memory reads/writes. As long as ctrulib provides these capabilities (i.e. `gfxInit` – and not only `gfxInitDefault` — as well as `gfxSetScreenFormat`) these functions should not cause unexpected behavior.

profi200:
My vote is on reallocating the framebuffers.

fincs:
Reallocate if current framebuffer is not big enough, that should be the proper way to fix this.

esoteric-programmer:
I've modified the pull request. I have not add additional variable to store the maximum size of the framebuffer, but realloc the framebuffer each time the size needed changes. Is it okay? Or should its maximal size be stored and only reallocated if the maximum size is not large enough?

fincs:
> Or should its maximal size be stored and only reallocated if the maximum size is not large enough?

This is what I suggested in my previous comment.

esoteric-programmer:
Sorry.

fox8091:
Could someone please at least make a comment on this? The last commit was almost a month ago and there are *4* open PRs with no feedback. Close and tell why or pull or give feedback on how to fix it or something. This is ridiculous. No feedback whatsoever.

PR #323, opened 2016-12-18
Change svcKernelSetStateSignature

TuxSH:
`svcKernelSetState` behaves like a variadic function. This change, in addition of not breaking anything, would allow writing things like:

```c
svcKernelSetState(0, 0x0004013800000002ULL, 0); // firm-launch
svcKernelSetState(7); // MCU reboot
````

which is more explicit than what was used before


TuxSH:
Damn, I made a typo X)

PR #325, opened 2016-12-21
Finish implementing the debug SVCs, fix bugs, other changes

TuxSH:
This completes the definition of `DebugEventInfo` and defines the remaining types for the debug SVCs, implement the remaining SVCs, etc.

### DIFF
--- a/libctru/include/3ds/errf.h
+++ b/libctru/include/3ds/errf.h
@@ -7,11 +7,6 @@
 
 #include <3ds/types.h>
 
-/// Used for register dumps.
-typedef struct {
-	u32 r0, r1, r2, r3, r4, r5, r6, r7, r8, r9, r10, r11, r12, r13, r14, r15, cpsr;
-} ERRF_ExceptionContext;
-
 /// Types of errors that can be thrown by err:f.
 typedef enum {
 	ERRF_ERRTYPE_GENERIC      = 0, ///< For generic errors. Shows miscellaneous info.
@@ -19,7 +14,7 @@ typedef enum {
 	ERRF_ERRTYPE_CARD_REMOVED = 2, ///< Displays the "The Game Card was removed." message.
 	ERRF_ERRTYPE_EXCEPTION    = 3, ///< For exceptions, or more specifically 'crashes'. union data should be exception_data.
 	ERRF_ERRTYPE_FAILURE      = 4, ///< For general failure. Shows a message. union data should have a string set in failure_mesg
-	ERRF_ERRTYPE_LOGGED       = 5  ///< Outputs logs to NAND in some cases.
+	ERRF_ERRTYPE_LOGGED       = 5, ///< Outputs logs to NAND in some cases.
 } ERRF_ErrType;
 
 /// Types of 'Exceptions' thrown for ERRF_ERRTYPE_EXCEPTION
@@ -27,7 +22,7 @@ typedef enum {
 	ERRF_EXCEPTION_PREFETCH_ABORT = 0, ///< Prefetch Abort
 	ERRF_EXCEPTION_DATA_ABORT     = 1, ///< Data abort
 	ERRF_EXCEPTION_UNDEFINED      = 2, ///< Undefined instruction
-	ERRF_EXCEPTION_VFP            = 3  ///< VFP (floating point) exception.
+	ERRF_EXCEPTION_VFP            = 3, ///< VFP (floating point) exception.
 } ERRF_ExceptionType;
 
 typedef struct {
@@ -42,7 +37,7 @@ typedef struct {
 
 typedef struct {
 	ERRF_ExceptionInfo excep;   ///< Exception info struct
-	ERRF_ExceptionContext regs; ///< Register dump.
+	CpuRegisters regs;          ///< CPU register dump.
 	u8 pad[4];
 } ERRF_ExceptionData;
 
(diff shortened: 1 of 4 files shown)


libctru/include/3ds/svc.h
```diff
@@ -930,11 +988,19 @@ Result svcKernelSetState(u32 type, ...);
...
+void svcDebugBreak(UserBreakType breakReason, u32 debugParameter1, u32 debugParameter2);
+
+/**
  * @brief Outputs a debug string.
  * @param str String to output.
- * @param length Length of the string to output.
+ * @param length Length of the string to output (needs to be positive).
  */
-Result svcOutputDebugString(const char* str, int length);
+Result svcOutputDebugString(const char* str, s32 length);
```
mtheall: Can this be a size_t or u32?

TuxSH: So it'd be `ssize_t` for this case. However the rest of the SVCs prototypes are using `u32` for size parameters.

fincs: I'm curious, what is the significance of a negative value when passed to this parameter?

fincs: Then I don't see why you would want to make it a signed integer then ( ͡° ͜ʖ ͡°)

TuxSH: Because the kernel uses signed comparisons on it (BLT)? ( ͡° ͜ʖ ͡°)

fincs: That doesn't mean anything from the user's point of view. Signed types exist in order to allow for negative numbers. Since negative numbers are treated as "invalid arguments out of range", it makes no sense to allow for them.

profi200: If the sign bit is treated as flag i would make it unsigned. Makes it easier to deal with it imho.

TuxSH: To recap, `OutputDebugString` treats the string's size as a signed parameter. If it is negative, an error is returned. There is code that extracts the sign bit of the size parameter (bit which is thus always 0)

mtheall: Signed integers don't have a "sign bit" per se. It just happens to be that the high bit of a negative number is always set, but that's just a side-effect of twos-complement form. If they truly had a sign bit, you would gain -0 but lose -2147483648.

In any case, the kernel treats this parameter as signed. I would therefore leave the size parameter as s32.

TuxSH: Of course

libctru/include/3ds/svc.h
```diff
@@ -994,7 +994,7 @@ void svcBreak(UserBreakType breakReason);
  * @param croInfo Library information.
  * @param croInfoSize Size of the above structure.
  */
-void svcBreakRO(UserBreakType breakReason, const void* croInfo, u32 croInfoSize);
+void svcBreakRO(UserBreakType breakReason, const void* croInfo, u32 croInfoSize) __asm__("svcBreak");
```
mtheall: Could you instead do:

    void svcBreakRO(UserBreakType breakReason, const void* croInfo, u32 croInfoSize);
    static inline void svcBreak(UserBreakType breakReason)
    {
      return svcBreakRO(breakReason, NULL, 0);
    }

and implement svcBreakRO in libctru/source/svc.s

fincs: The syscall is not called svcBreakRO and the extra parameters do not exist (not even dummy NULLs) for non-RO break reasons (syscall ABI is *not* ARM-EABI).

TuxSH: Not sure it'll be beneficial: current homebrew and official software use the "default" version of svcBreak with one argument...

The non-default version of svcBreak is only used for debugging purposes.

libctru/include/3ds/errf.h
```diff
@@ -15,15 +15,15 @@ typedef enum {
...
+} PACKED ERRF_ErrType;
 
 /// Types of 'Exceptions' thrown for ERRF_ERRTYPE_EXCEPTION
 typedef enum {
 	ERRF_EXCEPTION_PREFETCH_ABORT = 0, ///< Prefetch Abort
 	ERRF_EXCEPTION_DATA_ABORT     = 1, ///< Data abort
 	ERRF_EXCEPTION_UNDEFINED      = 2, ///< Undefined instruction
 	ERRF_EXCEPTION_VFP            = 3, ///< VFP (floating point) exception.
-} ERRF_ExceptionType;
+} PACKED ERRF_ExceptionType;
```
WinterMute: This PACKED thing isn't a bugfix, it's telling the compiler to use the minimum size to contain all the enum values. This is default on arm-none-eabi anyway so I'm not entirely sure what we gain from this,

This should be it's own commit detailing what it does and why.

libctru/source/svc.s
```diff
@@ -421,16 +421,12 @@ SVC_BEGIN svcContinueDebugEvent
 	bx  lr
 
 SVC_BEGIN svcGetProcessList
-	push {r0, r1}
+	str r0, [sp, #-0x4]!
```
WinterMute: again, commit this code separately - you can use -m twice to give a title & a summary,

i.e. git commit -m "fix svcGetProcessList" -m "Fixing the implementation of svcGetProcessList - one output parameter not two."

TuxSH:
@WinterMute I think you just need to browse the commit -- maybe things broke because of me force-pushing

PR #326, opened 2016-12-28
Add APT:ReceiveDeliverArg

endrift:
APT:ReceiveDeliverArg is used to retrieve the "deliver" arguments, which include the parameter buffer passed to APT:DoApplicationJump.

### DIFF
--- a/libctru/source/services/apt.c
+++ b/libctru/source/services/apt.c
@@ -1079,3 +1079,36 @@ Result APT_GetSharedFont(Handle* fontHandle, u32* mapAddr)
 
 	return ret;
 }
+
+Result APT_ReceiveDeliverArg(const void* param, size_t paramSize, const void* hmac, u64* sender, bool* received)
+{
+	u32 cmdbuf[16];
+	cmdbuf[0]=IPC_MakeHeader(0x35,2,0); // 0x350080
+	cmdbuf[1]=paramSize;
+	cmdbuf[2]=hmac ? 0x20 : 0;
+
+	u32 saved_threadstorage[4];
+	u32* staticbufs = getThreadStaticBuffers();
+	saved_threadstorage[0]=staticbufs[0];
+	saved_threadstorage[1]=staticbufs[1];
+	saved_threadstorage[2]=staticbufs[2];
+	saved_threadstorage[3]=staticbufs[3];
+	staticbufs[0]=IPC_Desc_StaticBuffer(cmdbuf[1],0);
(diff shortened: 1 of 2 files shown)


libctru/source/services/apt.c
```diff
@@ -1079,3 +1079,36 @@ Result APT_GetSharedFont(Handle* fontHandle, u32* mapAddr)
...
+
+	Result ret = aptSendCommand(cmdbuf);
+	staticbufs[0]=saved_threadstorage[0];
+	staticbufs[1]=saved_threadstorage[1];
+	staticbufs[2]=saved_threadstorage[2];
+	staticbufs[3]=saved_threadstorage[3];
+
+	if (R_SUCCEEDED(ret))
+	{
+		if (sender)     *sender    =cmdbuf[2];
```
fincs: u64 output parameter, however you are assigning a single 32-bit value to it? Please check if this is correct.

libctru/source/services/apt.c
```diff
@@ -1079,3 +1079,36 @@ Result APT_GetSharedFont(Handle* fontHandle, u32* mapAddr)
...
+		if (received)   *received  =cmdbuf[3];
```
fincs: For consistency's sake, make this `cmdbuf[3] & 0xFF`.

libctru/source/services/apt.c
mtheall: See https://www.3dbrew.org/wiki/APT:ReceiveDeliverArg
This output spans cmdbuf[2-3].

    if(sender) *sender = *((u64*)&cmdbuf[2]);

or

    // Alignment safe?
    if(sender) memcpy(sender, &cmdbuf[2], sizeof(*sender));

libctru/source/services/apt.c
mtheall: See https://www.3dbrew.org/wiki/APT:ReceiveDeliverArg
This output is in cmdbuf[4].

    if (received) *received = !!(cmdbuf[4]); // Convert to a real bool

libctru/source/services/apt.c
fincs: Why not
```c
if(sender) *sender = cmdbuf[2] | ((u64)cmdbuf[3]<<32);
```
?

libctru/source/services/apt.c
Lectem: >      !!(cmdbuf[4]); // Convert to a real bool

Don't forget the 0xFF mask, as per nintendo ABI

PR #331, opened 2017-01-16
Add nim:s client implementation.

AriA99:
This pull request allows usage of nim:s. I hope you find it useful.

### DIFF
--- a/libctru/include/3ds/services/nim.h
+++ b/libctru/include/3ds/services/nim.h
@@ -20,7 +20,7 @@ typedef enum
 	IM_DEFAULT = 0, ///< Initial installation
 	IM_UNKNOWN1,    ///< Unknown
 	IM_UNKNOWN2,    ///< Unknown
-	IM_REINSTALL    ///< Reinstall currently installed title; use this if the title is already installed (including updates)
+	IM_REINSTALL,   ///< Reinstall currently installed title; use this if the title is already installed (including updates)
 } NIM_InstallationMode;
 
 /// Current state of a NIM download/installation.
@@ -37,7 +37,7 @@ typedef enum
 	DS_VERSION_ERROR,       ///< (unknown error regarding title version)
 	DS_CREATE_CONTEXT,      ///< Creating the .ctx file?
 	DS_CANNOT_RECOVER,      ///< Irrecoverable error encountered (e.g. out of space)
-	DS_INVALID              ///< Invalid state
+	DS_INVALID,             ///< Invalid state
 } NIM_DownloadState;
 
 /// Input configuration for NIM download/installation tasks.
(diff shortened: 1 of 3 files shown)


libctru/include/3ds/services/nim.h
```diff
@@ -0,0 +1,132 @@
...
+// FS_MediaType
+#include <3ds/services/fs.h>
+
+/// Mode that NIM downloads/installs a title with.
+typedef enum
+{
+	IM_DEFAULT = 0, ///< Initial installation
+	IM_UNKNOWN1,    ///< Unknown
+	IM_UNKNOWN2,    ///< Unknown
+	IM_REINSTALL    ///< Reinstall currently installed title; use this if the title is already installed (including updates)
```
fincs: Needs trailing comma even in the last enum value.

libctru/include/3ds/services/nim.h
```diff
@@ -0,0 +1,132 @@
...
+	DS_DOWNLOAD_TMD,        ///< Downloading and installing TMD
+	DS_PREPARE_SAVE_DATA,   ///< Initializing save data
+	DS_DOWNLOAD_CONTENTS,   ///< Downloading and installing contents
+	DS_WAIT_COMMIT,         ///< Waiting before calling AM_CommitImportTitles
+	DS_COMMITTING,          ///< Running AM_CommitImportTitles
+	DS_FINISHED,            ///< Title installation finished
+	DS_VERSION_ERROR,       ///< (unknown error regarding title version)
+	DS_CREATE_CONTEXT,      ///< Creating the .ctx file?
+	DS_CANNOT_RECOVER,      ///< Irrecoverable error encountered (e.g. out of space)
+	DS_INVALID              ///< Invalid state
```
fincs: Same thing

libctru/source/services/nim.c
```diff
@@ -0,0 +1,254 @@
...
+	if (R_FAILED(ret = svcSendSyncRequest(nimsHandle))) return ret;
+
+	return (Result)cmdbuf[1];
+}
+
+static Result NIMS_SetTIN(void)
+{
+	Result ret = 0;
+	u32 *cmdbuf = getThreadCommandBuffer();
+	static char attr[] = "TIN";
```
fincs: Should be `static const char` in order to place it in rodata. Also, is there a reason why this is hardcoded instead of being a parameter? If necessary, add a wrapper function in the header.

libctru/source/services/nim.c
```diff
@@ -0,0 +1,254 @@
...
+	static char TIN[] = "56789"; // Obtained from eShop
```
fincs: As above.

libctru/source/services/nim.c
```diff
@@ -0,0 +1,254 @@
...
+	if (R_FAILED(ret = svcSendSyncRequest(nimsHandle))) return ret;
+
+	return (Result)cmdbuf[1];
+}
+
+static Result NIMS_ConnectNoTicketDownload(void)
+{
+	Result ret = 0;
+	u32 *cmdbuf = getThreadCommandBuffer();
+	static u8 buffer[0x20000];
```
fincs: Why is this hardcoded? Also, this is not thread safe. If you really do not want to allow the user to pass their own buffer, this should be marked as thread-local (refer to the sdmc or romfs code for an example).

libctru/source/services/nim.c
```diff
@@ -0,0 +1,254 @@
...
+	cfg->version = version;
+	cfg->ratingAge = ratingAge;
+	cfg->mediaType = mediaType;
+}
+
+Result NIMS_RegisterTask(const NIM_TitleConfig *cfg, const char *name, const char *maker)
+{
+	Result ret = 0;
+	u32 *cmdbuf = getThreadCommandBuffer();
+	static uint16_t name16[0x92], maker16[0x4a];
```
fincs: This should be non-static (on the stack).

libctru/source/services/nim.c
AriA99: This is hardcoded because I don't have enough data to not hardcode it. All I know is that the eShop passes a 0x20000 buffer in .bss to nim:s for this. It's unclear how big the buffer has to be. I did this in the name of user-friendliness but I can change it to take a user-provided buffer, adding a comment that the input buffer must be at least 0x20000 bytes.

Do you want the fixes to your comments as separate commit or rebased and force-pushed?

fincs: Just commit and when the PR is ready, I'll squash&merge everything using GitHub's button.

libctru/source/services/nim.c
AriA99: This is hardcoded because `nim:s` is only used by tiger, so only one TIN value is ever valid in this scenario. I just tested it and `nim` *does* accept other TINs, even if they're the wrong ones, so I can make this a parameter and have a user-friendly wrapper with this default instead.

yellows8: A wrapper would be better, this sounds like a general set-attribute command or so.

AriA99:
@fincs I believe I've addressed your requests and I am requesting a second review.

fincs:
LGTM.

WinterMute:
Looks fine. Squash when merging :)

PR #333, opened 2017-01-20
change members name

Slimfr01:
Change name of members of NFC_Amiiboconfig after some tests to identify amiibos.

### DIFF
--- a/libctru/include/3ds/services/nfc.h
+++ b/libctru/include/3ds/services/nfc.h
@@ -71,10 +71,10 @@ typedef struct {
 	u8 lastwritedate_month;
 	u8 lastwritedate_day;
 	u16 write_counter;
-	u16 val_x6;
-	u8 val_x8;
-	u8 val_x9;
-	u16 val_xa;
+	u8 characterID[3];
+	u8 serie;///ID of the serie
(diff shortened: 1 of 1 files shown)


mtheall:
This changes the size of the struct. The affected fields go from 6 bytes total to 7 bytes. Due to alignment of the new field 'type', this actually shifts the rest of the fields by 2 bytes. Are you sure this is correct?

Slimfr01:
In fact yes i made some mistake
After some test and corrections i manage to found the correct values for all the known data.

Digby Figure :
image

Digby Card ( Same info except amiiboID and type ) :
image

Hammer Slam Bowser ( different series, different character, variant ) :
image

for reference but note that i switch the term series to follow the official name ( Smash bros series, Animal Crossing series ... ) : https://docs.google.com/spreadsheets/d/1WJ4HxS9hkLquq-ATt1Rq9mioH6RDgP3qQrtYVaOdimM/edit#gid=628338232

It seem that all is good with this version that i will commit right away.

yellows8:
Replace val_xc with the pagex4_byte3 line which you removed. zeros[0x31] needs changed to zeros[0x30] since you changed appdata_size to u16.

PR #336, opened 2017-02-11
update ticks per msec/usec

FenrirWolf:
I noticed that system time functions seem to be measuring seconds as slightly longer than they actually are. This change brings the ticks per msec and usec values in line with the clock speed reported on https://www.3dbrew.org/wiki/Hardware#Common_hardware

### DIFF
--- a/libctru/source/os.c
+++ b/libctru/source/os.c
@@ -8,8 +8,8 @@
 #include <reent.h>
 #include <unistd.h>
 
-#define TICKS_PER_USEC 268.123480
-#define TICKS_PER_MSEC 268123.480
+#define TICKS_PER_USEC 268.111856
+#define TICKS_PER_MSEC 268111.856
 
 // Work around the VFP not supporting 64-bit integer <--> floating point conversion
 static inline double u64_to_double(u64 value) {


mtheall:
I see that TuxSH updated the page to change the frequency.

https://www.3dbrew.org/w/index.php?title=Hardware&type=revision&diff=18793&oldid=18015

Can anyone show how we know what the correct value is?

FenrirWolf:
I noticed the discrepancy when comparing the time before and after a call to svcSleepThread, so maybe that's a metric that can be used? https://github.com/rust3ds/ctru-rs/pull/18#issuecomment-279101989 for more context.

TuxSH:
@mtheall My value comes from kernel11 itself

See https://github.com/citra-emu/citra/issues/2445.

WinterMute:
Yeah. That's not right - you're looking at what a compiler has turned a calculation into combined with a constant that's more than likely calculated at compile time.

On phone atm. I'll have a better look later

WinterMute:
@Fe

TuxSH:
ns_high@r4, ns_low@r5

```asm
LDR             R0, =0x44A2FA85
MOV             R1, R4,ASR#31
MOV             R2, #0
LDR             R7, =TimerAndWDTManager
UMULL           R5, R12, R5, R0
UMULL           R5, R8, R4, R0
MLA             R0, R1, R0, R8
MLA             R0, R4, R2, R0
ADDS            R4, R5, R12
ADC             R5, R0, R2
MOV             R0, R7  ; this
BL              KTimerAndWDTManager__GetSystemTick
ADDS            R4, R4, R0
ADC             R5, R5, R1
MOV             R0, #0
MOV             R1, R0
SUBS            R0, R0, R4
SBCS            R0, R1, R5
MOVGE           R4, #0xFFFFFFFF
MOVGE           R5, #0x7FFFFFFF
```

TuxSH:
Nothing is done is timeout < 0, and

```c
    if ( nanoSeconds > 0x9A2 )
    {
      ns_low = nanoSeconds - 0x9A2;
      ns_high = HIDWORD(nanoSeconds) - !__CFADD__((_DWORD)nanoSeconds, 0xFFFFF65E);
    }
    else
    {
      ns_low = 1;
      ns_high = 0;
    }````

or if you prefer:

```asm
LDR             R1, =0x9A2
SUBS            R1, R1, R2
RSCS            R0, R3, #0
MOVGE           R5, #1
MOVGE           R4, #0
BGE             loc_FFF20720

LDR             R0, =-0x9A2
ADDS            R5, R2, R0
SBC             R4, R3, #0

loc_FFF20720:```

FenrirWolf:
@WinterMute I also supposed rounding errors were at fault at first, but my testing leads me to believe that the clock speed @TuxSH came up with is the right one, or at least is closer to the right value. Here's a few tests I did compiling my code against libctru with the old constant vs the new one. The code was run on my n3ds in the homebrew launcher:

#define TICKS_PER_USEC 268.123480:

Sleep for 1 second: `SystemTime: Duration { secs: 0, nanos: 999989000 }`
Sleep for 2 seconds: `SystemTime: Duration { secs: 1, nanos: 999950000 }`
Sleep for 3 seconds: `SystemTime: Duration { secs: 2, nanos: 999902000 }`
Sleep for 4 seconds: `SystemTime: Duration { secs: 3, nanos: 999859000 }`
Sleep for 5 seconds: `SystemTime: Duration { secs: 4, nanos: 999815000 }`

#define TICKS_PER_USEC 268.111855:

Sleep for 1 second: `SystemTime: Duration { secs: 1, nanos: 33000 }`
Sleep for 2 seconds: `SystemTime: Duration { secs: 2, nanos: 33000 }`
Sleep for 3 seconds: `SystemTime: Duration { secs: 3, nanos: 35000 }`
Sleep for 4 seconds: `SystemTime: Duration { secs: 4, nanos: 35000 }`
Sleep for 5 seconds: `SystemTime: Duration { secs: 2, nanos: 34000 }`

These are only two example runs, but I did each test multiple times and the results were similar each time. The thing that stands out the most to me is the linearly decreasing number of nanoseconds in the first run, which is consistent with the notion of each "second" reported by `gettimeofday()` being slower than the number of seconds that `svcSleepThread` is actually sleeping for.

Meanwhile in the second run the number of nanoseconds elapsed stays pretty firmly in the 30000-40000 range regardless of the number of seconds elapsed.

WinterMute:
We know the bus clock on the DS is 33513982 Hz - http://problemkaputt.de/gbatek.htm#dsmemorytimings

The DS ARM9 is clocked at 2x bus clock, or 67027964Hz and the DSi ARM9 is double again at 134055928Hz.

It doesn't take a huge leap of imagination to decide that the 3DS ARM11 might be clocked at double the DSi ARM9 clock at 268111856Hz. Obviously it's easier to use multiples of the same base clock for each of the systems built into the 3DS.

((1<<32)*268111856)/10⁹ = 1151531653.189861376 which gives you your constant of 0x44A2FA85

Given the consistency of @FenrirWolf's tests I'd be happy with a change to 268111856

Somehow I expected a larger range of possible values for the clock to produce that constant but there you go. The math doesn't lie.

FenrirWolf:
Should I force-push a change to 268111.856 or make a new commit?

WinterMute:
Yeah, just force push

FenrirWolf:
Done.

PR #344, opened 2017-03-02
 Return zero when writing zero bytes to the console

FenrirWolf:
Writing zero bytes to stdout just does nothing and doesn't set errno, so it shouldn't return -1.

### DIFF
--- a/libctru/source/console.c
+++ b/libctru/source/console.c
@@ -187,7 +187,7 @@ ssize_t con_write(struct _reent *r,void *fd,const char *ptr, size_t len) {
 	int i, count = 0;
 	char *tmp = (char*)ptr;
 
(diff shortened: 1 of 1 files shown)


libctru/source/console.c
```diff
@@ -187,7 +187,7 @@ ssize_t con_write(struct _reent *r,void *fd,const char *ptr, size_t len) {
 	int i, count = 0;
 	char *tmp = (char*)ptr;
 
-	if(!tmp || len<=0) return -1;
+	if(!tmp || len<0) return -1;
```
mtheall: `len<0` is always false since `len` is unsigned. That would simplify this to `if(!tmp) return -1`.

FenrirWolf: derp, good catch. fixed.

PR #351, opened 2017-04-03
Implement GSPLCD::GetBrightness

joel16:
I've implemented this using the information given on https://www.3dbrew.org/wiki/GSPLCD:GetBrightness

From what I've tested these are the brightness values that I have obtained:

Brightness 1 = 0x10
Brightness 2 = 0x1c
Brightness 3 = 0x30
Brightness 4 = 0x52
Brightness 5 = 0x8e

### DIFF
--- a/libctru/include/3ds/services/gsplcd.h
+++ b/libctru/include/3ds/services/gsplcd.h
@@ -36,3 +36,9 @@ Result GSPLCD_PowerOffBacklight(u32 screen);
  * @param vendor Pointer to output the screen vendors to.
  */
 Result GSPLCD_GetVendors(u8 *vendors);
+
+/**
+ * @brief Gets the LCD screens' brightness. Stubbed on old 3ds.
+ * @param screen to get the brightness value.
+ */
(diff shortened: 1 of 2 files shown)


mtheall:
The input and output are different"types". I would do something like this:

`Result GSPLCD_GetBrightness(u32 screen, u32 *brightness);`

joel16:
Done, tested and confirmed working. Please let me know if I need to do anything else

mtheall:
Looks good to me. The only improvements I can think of are checking for `brightness` being non-null and using an enum for `screen`.

Enum for `brightness` might make sense, but I wonder how "power saving" mode, having 3D on/off, hardware unit (2DS/3DS/New 3DS, XL or not), and screen type (TN/IPS) affect the values. And there's always the fact that the max value you encountered is 0x8E where 3dbrew wiki suggests a max value of 0xAC.

Maybe it's simply best to have `brightness` be a u8.

joel16:
Thanks for the merge.
I'll look into your suggestion once I get some free time.

PR #353, opened 2017-05-13
Add support for user-specified exception handlers

TuxSH:
This pull request add support for user-specified exception handlers. It also provides a ready-to-use exception handler which uses ErrDisp, `ERRF_ExceptionHandler` (you'll however need to make sure that ENVINFO's bit0 is clear to be able to see any information, if using the official ErrDisp implementation).

It is needed for UNITNFO to be non-zero on Kernel11 start, for CPU exceptions to be reported this way (one could also have the relevant checks patched).

Example code:
```c
	/* Enable the reporting of CPU exception to user-specified handlers
	   One could also set bit0 of the byte at offset 0xC in the 0x000F0000 configuration block
	*/
	svcKernelSetState(6, 1, 1LL);

	threadOnException(ERRF_ExceptionHandler, RUN_HANDLER_ON_FAULTING_STACK, WRITE_DATA_TO_HANDLER_STACK);
	__asm__ volatile("bkpt 1");
```

### DIFF
--- a/libctru/include/3ds/errf.h
+++ b/libctru/include/3ds/errf.h
@@ -110,3 +110,13 @@ Result ERRF_ThrowResult(Result failure);
  * on development units/patched ErrDisp.
  */
 Result ERRF_ThrowResultWithMessage(Result failure, const char* message);
+
+/**
+ * @brief Handles an exception using ErrDisp.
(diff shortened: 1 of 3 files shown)


fincs:
:lgtm:

PR #355, opened 2017-05-21
Fix compilation issue caused due to errorConf.

Pirater12:


fincs:
Can you rename it to `type` (in lowercase)?

PR #358, opened 2017-07-02
nfcStartOtherTagScanning(): Call StartCommunication() for Old3DS compatibility.

HiddenRambler:
Duplicated Old 3DS compatibility code from nfcStartTagScanning() for nfcStartOtherTagScanning(). Could refactor more if needed.

### DIFF
--- a/libctru/source/services/nfc.c
+++ b/libctru/source/services/nfc.c
@@ -15,6 +15,7 @@ static NFC_OpType nfc_optype = NFC_OpType_NFCTag;
 static Result NFC_Initialize(NFC_OpType type);
 static Result NFC_Shutdown(NFC_OpType type);
 
+static Result NFC_StartCommsAdapter();
 static Result NFC_StartCommunication(void);
 static Result NFC_StopCommunication(void);
 static Result NFC_CommunicationGetStatus(u8 *out);
@@ -64,7 +65,7 @@ Handle nfcGetSessionHandle(void)
 	return nfcHandle;
 }
 
-Result nfcStartScanning(u16 inval)
+static Result NFC_StartCommsAdapter()
 {
 	Result ret, ret2;
 	bool new3ds_flag = false;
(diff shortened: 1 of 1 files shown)


fincs:
Instead of duplicating the code, why not extract it out to a separate static function?

HiddenRambler:
Refactored the shared code out.

libctru/source/services/nfc.c
```diff
@@ -15,12 +15,14 @@ static NFC_OpType nfc_optype = NFC_OpType_NFCTag;
 static Result NFC_Initialize(NFC_OpType type);
 static Result NFC_Shutdown(NFC_OpType type);
 
+static Result NFC_StartCommsAdapter();
```
fincs: Should be `(void)` just like the others - in C, `()` doesn't indicate an empty parameter list.

libctru/source/services/nfc.c
```diff
@@ -63,49 +65,62 @@ Handle nfcGetSessionHandle(void)
 	return nfcHandle;
 }
 
-Result nfcStartScanning(u16 inval)
+static Result NFC_StartCommsAdapter()
```
fincs: Same thing here

HiddenRambler:
Done.

WinterMute:
What's the difference between StartOtherTagScanning and StartTagScanning here? From a brief glance it looks to me like the former is a superset of the latter and perhaps they should be only differentiated internally?

Would it be reasonable to provide a single interface which subsequently calls the appropriate function based on it's parameters? 

@yellows8

HiddenRambler:
This would depend on how much the ctrulib will abstract away the native api.

While we are at it. According to my trial and error investigation the nfcCmd21() and nfcCmd22() seems to enable and disable keeping the NFC tag powered up during commands. This is usefull to perform as part of a single session of the tag. see: https://github.com/HiddenRambler/Thenaya/blob/master/source/nfc.c#L332

perhaps we can rename these methods to more sensible names?

WinterMute:
Merging this in as is for next release.

Abstracting out the NFC api more can only be a good thing I think and better names for some of the partially investigated functions would be appreciated.

We can look at that for another PR though rather than leaving bugfixes sitting in review.

yellows8:
"What's the difference between StartOtherTagScanning and StartTagScanning here?" Basically enables using this: https://3dbrew.org/wiki/NFC:SendTagCommand

WinterMute:
@yellows8 ok, is there any reason why we would want to use the more limited version in preference?

yellows8:
Whenever one wants to send raw NFC commands.

PR #363, opened 2017-07-14
Implement GSPLCD_SetBrightness and GSPLCD_SetBrightnessRaw

joel16:
Implemented both GSPLCD_SetBrightness and GSPLCD_SetBrightnessRaw.
Thanks to the info from:

https://www.3dbrew.org/wiki/GSPLCD:SetBrightness
https://www.3dbrew.org/wiki/GSPLCD:SetBrightnessRaw

### DIFF
--- a/libctru/include/3ds/services/gsplcd.h
+++ b/libctru/include/3ds/services/gsplcd.h
@@ -50,3 +50,10 @@ Result GSPLCD_GetBrightness(u32 screen, u32 *brightness);
  * @param brightness Brightness value set.
  */
 Result GSPLCD_SetBrightness(u32 screen, u32 brightness);
+
+/**
+ * @brief Sets the LCD screens' raw brightness.
+ * @param screen Screen to set the brightness value of.
+ * @param brightness Brightness value set.
+ */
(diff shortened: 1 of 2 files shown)


Pirater12:
@joel16 I think you should mention in the header file that top screen = 1 and bottom screen = 2

joel16:
I mean I would, but I'll wait for either fincs or wintermute or mtheall to make a comment first. Simply because the other function definitions doesn't mention the values of screen index.

Also it seems these enums won't work:

GSPLCD_SCREEN_TOP    = BIT(GFX_TOP) // 0
GSPLCD_SCREEN_BOTTOM = BIT(GFX_BOTTOM) // 1

So I was wondering if another enum would be better perhaps something like,

GSPLCD_BRIGHTNESS_LCD_TOP
GSPLCD_BRIGHTNESS_LCD_BOTTOM

I'll wait for an official word though.

WinterMute:
Handily enough   GSPLCD_SCREEN_TOP is 1 and GSPLCD_SCREEN_BOTTOM is 2 :p

PR #364, opened 2017-07-27
Bump most recent version

ahoischen:
Updates the "most recent version at the time of this writing" to r47.


WinterMute:
bumping the version number for every devkitARM release will just cause unnecessary noise in the repository. For that reason I'm removing the text to remove further temptation.

PR #366, opened 2017-07-30
Add exheader.h

TuxSH:
This adds NCCH extended header definitions, required by `pm`, `loader`, PXIPM and fs:REG commands that I'll implement later.


fincs:
You should add an include line to 3ds.h.

PR #372, opened 2017-08-08
Add Mii selector applet (appletEd)

phijor:
Lately, @s0r00t and I have been reverse engineering the Mii Selector applet. This is a working implementation of how to configure and launch it. I documented what we found out [in the wiki](https://www.3dbrew.org/wiki/Mii_Selector).

Please criticize any of the code and/or documentation. I tried to keep the implementation close to how the interface to `swkbd` and `errDisp` is implemented, but right now this code is far less verbose than that of those applets.

### DIFF
--- a/libctru/include/3ds/applets/miiselector.h
+++ b/libctru/include/3ds/applets/miiselector.h
@@ -108,4 +108,4 @@ Result miiSelectorLaunch(MiiSelectorContext *ctx);
  *
  * @return `true` if `returnbuf->checksum` is the same as the one computed from `returnbuf`
  */
-bool miiSelectorChecksumIsValid(MiiSelectorReturn *returnbuf);
+bool miiSelectorChecksumIsValid(const MiiSelectorReturn *returnbuf);
--- a/libctru/include/3ds/applets/miiselector.h
+++ b/libctru/include/3ds/applets/miiselector.h
@@ -5,11 +5,6 @@
 
 #pragma once
 
-/// @cond INTERNAL
-#define _assert_struct_size(s, size) \
-	_Static_assert(sizeof(s) == (size), #s " is not of size " #size "!")
-/// @endcond
-
 /// Magic value needed to launch the applet.
 #define MIISELECTOR_MAGIC 0x13DE28CF
 
@@ -51,9 +46,6 @@ typedef struct
 	           ///< applet.
 } MiiSelectorConf;
 
-/// @private
-_assert_struct_size(MiiSelectorConf, 0x104);
-
 /// Size of the data describing a single Mii
 #define MIISELECTOR_MIIDATA_SIZE 0x5c
 
@@ -78,11 +70,6 @@ typedef struct
 	                                                   ///< if one was selected (UTF16-LE
 	                                                   ///< string). Zeroed otherwise.
 } MiiSelectorReturn;
-
-/// @private
-_assert_struct_size(MiiSelectorReturn, 0x84);
-
-#undef _assert_struct_size
 
 /**
  * @brief Launch the Mii selector library applet
--- a/libctru/include/3ds/applets/miiselector.h
+++ b/libctru/include/3ds/applets/miiselector.h
@@ -85,20 +85,12 @@ _assert_struct_size(MiiSelectorReturn, 0x84);
 #undef _assert_struct_size
 
 /**
- * @brief Mii selector context
- *
- * Configure how to launch the applet in @ref MiiSelectorContext.config. After
- * returning, the applet has written data to @ref MiiSelectorContext.ret.
- */
-typedef union {
-	MiiSelectorConf config; ///< Configuration passed to applet
-	MiiSelectorReturn ret;  ///< Data written by applet after return
-} MiiSelectorContext;
-
-/**
  * @brief Launch the Mii selector library applet
+ *
+ * @param conf Configuration determining how the applet should behave
+ * @param returnbuf Data returned by the applet
  */
-Result miiSelectorLaunch(MiiSelectorContext *ctx);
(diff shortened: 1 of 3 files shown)


s0r00t:
Making a usage example could be a swift addition. 
Also, why didn't you use booleans instead of chars in the struct for flags?

libctru/include/3ds/applets/miiselector.h
```diff
@@ -0,0 +1,111 @@
+/**
+ * @file miiselector.h
+ * @brief Mii Selector Applet (appletEd).
+ */
+
+#pragma once
+
+/// @cond INTERNAL
+#define _assert_struct_size(s, size) \
+	_Static_assert(sizeof(s) == (size), #s " is not of size " #size "!")
```
mtheall: This limits you to C11. I would do it like this:
```
/// @cond INTERNAL
#ifdef __cplusplus
#if __cplusplus >= 201103L
#define _assert_struct_size(s, size) \
	static_assert(sizeof(s) == (size), #s " is not of size " #size "!")
#endif
#else
#if __STDC_VERSION__ >= 201112L
#define _assert_struct_size(s, size) \
	_Static_assert(sizeof(s) == (size), #s " is not of size " #size "!")
#endif
#endif
/// @endcond

File: libctru/include/3ds/applets/miiselector.h [hunk 2]
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * @brief Mii selector context
+ *
+ * Configure how to launch the applet in @ref MiiSelectorContext.config. After
+ * returning, the applet has written data to @ref MiiSelectorContext.ret.
+ */
+typedef union {
+	MiiSelectorConf config; ///< Configuration passed to applet
+	MiiSelectorReturn ret;  ///< Data written by applet after return
+} MiiSelectorContext;
```
mtheall: Instead of using a union, I would have miiSelectorLaunch() take two parameters, one for the input MiiSelectorConf and one for the output MiiSelectorReturn.

libctru/include/3ds/applets/miiselector.h
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * @brief Launch the Mii selector library applet
+ */
+Result miiSelectorLaunch(MiiSelectorContext *ctx);
```
mtheall: From above, this becomes:

    Result miiSelectorLaunch(const MiiSelectorConf *conf, MiiSelectorReturn *returnbuf);

libctru/include/3ds/applets/miiselector.h
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * @brief Verifies that the Mii data returned from the applet matches its
+ * checksum
+ *
+ * @param returnbuffer Buffer filled by Mii selector applet
+ *
+ * @return `true` if `returnbuf->checksum` is the same as the one computed from `returnbuf`
+ */
+bool miiSelectorChecksumIsValid(MiiSelectorReturn *returnbuf);
```
mtheall: Use `const MiiSelectorReturn *returnbuf`

libctru/source/applets/miiselector.c
```diff
@@ -0,0 +1,42 @@
+#include <3ds/types.h>
+#include <3ds/services/apt.h>
+
+#include <3ds/applets/miiselector.h>
+
+Result miiSelectorLaunch(MiiSelectorContext *ctx)
```
mtheall: From above, this would look like the following:

```
Result miiSelectorLaunch(const MiiSelectorConf *conf, MiiSelectorReturn *returnbuf);
{
	union {
		MiiSelectorConf config;
		MiiSelectorReturn ret;
	} ctx;

	memcpy(&ctx.config, conf, sizeof(MiiSelectorConf));
	ctx.config.magic = MIISELECTOR_MAGIC;

	Result ret = aptLaunchLibraryApplet(APPID_APPLETED, &ctx->config, sizeof(MiiSelectorConf), 0);
	if(R_SUCCEEDED(ret) && returnbuf)
		memcpy(returnbuf, &ctx.ret, sizeof(MiiSelectorReturn));

	return ret;
}
```

phijor:
Adding an example to [devkitPro/3ds-examples ](https://github.com/devkitPro/3ds-examples) would be a good idea, indeed.

As it stands right now, I do not know if `sizeof(bool) == sizeof(u8)`. The structure critically depends on having a fixed memory layout, that's why I did not dare making the flags `bool`s.

libctru/include/3ds/applets/miiselector.h
fincs: Why is this here anyway? It doesn't belong in this header file (maybe types.h?)

libctru/source/applets/miiselector.c
phijor: I wasn't really happy with the `union` too; this is definitely a better approach. Thanks for the input, will implement!

libctru/include/3ds/applets/miiselector.h
phijor: @mtheall: Sorry, forgot that this was C11 only. What should this macro expand to if neither of your suggested conditions are met, `((void) 0)`?

@fincs: Of all the exisiting files I would guess `types.h` to be the most fitting, I don't really like here either (just didn't want to touch other files rn). Are there any other places in ctrulib that could benefit from using this macro or similar ones?

fincs: Can we get rid of this assert_struct_size thing? I don't think it's necessary when it's already verified to be correct.

phijor: Done.

PR #382, opened 2017-12-28
shaderProgram: correct array index in shaderProgramSetGshInputPermutation()

lioncash:


fincs:
Whoops, thank you for spotting this typo.

PR #384, opened 2018-01-03
gfx: Mark gfxSetFramebufferInfo and gfxWriteFramebufferInfo as static

lioncash:
These aren't exposed in the gfx header via a prototype or used anywhere else internally outside this translation unit, so I'm assuming these were intended to be static.

### DIFF
--- a/libctru/source/gfx.c
+++ b/libctru/source/gfx.c
@@ -102,7 +102,7 @@ void gfxSetDoubleBuffering(gfxScreen_t screen, bool doubleBuffering) {
 	doubleBuf[screen] = doubleBuffering ? 1 : 0; // make sure they're the integer values '1' and '0'
 }
 
-void gfxSetFramebufferInfo(gfxScreen_t screen, u8 id)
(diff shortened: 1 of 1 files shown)


fincs:
It looks like the fact that these functions aren't exposed in the header [hasn't stopped](https://github.com/shinyquagsire23/gpsp/blob/2ec0d851678922b4f57601117b942ca11e4ed0d9/3ds/_3ds.c#L150) [many different](https://github.com/FlexingTiger/Snes9xVITA/blob/dba30f374d6d4a4eb070257791d264ecf7294c7a/source/3dsgpu.cpp#L184-L185) people from using them, even though they weren't intended to be used :\

lioncash:
I agree in this case, since they're functions that weren't exposed in the public interface headers, so it was likely a conscious choice to deviate (or copy pasting from somewhere else that did it, which would be even worse).

fincs:
┬┴┬┴┤( ͡° ͜ʖ‌‌‌‌‌‌‌‌‌‌‌‌‌‌‌‌‌‌‌‌‌‌‌‌‌├┬┴┬┴

PR #392, opened 2018-01-28
Implement 8.3 filename (ShortName).

joel16:


fincs:
The data for short names was already inside `FS_DirectoryEntry`. Also, your change breaks said structure. What's the point of this?

joel16:
It was more or less for accuracy purposes, but if it breaks the DirectEntry struct feel free to close the PR.

PR #400, opened 2018-03-17
Initial bring up of the Friend Service

joel16:
Credits goes to those who documented these functions on 3Dbrew and Ryuzaki-MrL.

### DIFF
--- a/libctru/include/3ds/services/frd.h
+++ b/libctru/include/3ds/services/frd.h
@@ -1,0 +1,171 @@
+/**
+ * @file frd.h
+ * @brief Friend Services
+ */
+ #pragma once
+#include <3ds.h>
+
+#define FRIENDS_SCREEN_NAME_SIZE 0x16   // 11 (0x16 because UTF-16)
+#define FRIENDS_COMMENT_SIZE 0x22       // 16 (0x21 because UTF-16 + null character)
+#define FRIEND_LIST_SIZE 0x64           // 100 (Number of Friends)
+#define FRIEND_MII_STORE_DATA_SIZE 0x60 // 96 (Mii data)
+
+/// Friend key data
+typedef struct
+{
+   u32 principalId;
+   u32 padding;
+   u64 localFriendCode;
+} FriendKey;
+
+/// Structure containing basic Mii information.
+typedef struct 
+{
+    u32 mii_id;
+    u64 system_id;
+    u32 cdate;
+    u8 mac[0x6];
+    u16 padding;
+    u16 misc1;
+    u16 mii_name[0xB];
+    u8 width;
+    u8 height;
+    u32 misc2;
+    u32 unknown1;
+    u32 misc3;
+    u32 unknown2;
+    u8 allow_copy;
+    u8 unknown3[0x7];
+    u16 author[0xB];
+} MiiData;
+
+/// Friend profile data
+typedef struct
+{
+    u8 region;      // The region code for the hardware.
+    u8 country;     // Country code.
+    u8 area;        // Area code.
+    u8 language;    // Language code.
+    u8 platform;    // Platform code.
+    u32 padding;
+} Profile;
+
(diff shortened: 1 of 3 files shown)


libctru/include/3ds/services/frd.h
```diff
@@ -0,0 +1,171 @@
...
+    u32 unknown2;
+    u8 allow_copy;
+    u8 unknown3[0x7];
+    u16 author[0xB];
+} MiiData;
+
+/// Friend profile data
+typedef struct
+{
+    u8 region;      // The region code for the hardware.
```
Pirater12: You should probably change this to (CFG_Region)[https://github.com/smealum/ctrulib/blob/master/libctru/include/3ds/services/cfgu.h#L17]

fincs: Enums inside struct types that are going to be used for interfacing with external code is not exactly a good idea; because enum sizes are not defined by the standard and it is up to the compiler and ABI to decide which size they have by default. For instance, ARM (32-bit) defines they shall have the smallest size that will fit all values in the enum, but ARM (64-bit) defines them to have at least the size of int.

Pirater12: Oh okay

fincs:
On a closer inspection, I noticed that it is using UTF-16 strings as input or output. For consistency's sake, I'd suggest to implement transparent conversion to/from UTF-8, also to make the lives of users of this API easier. Note that libctru has built-in utf8/utf16 conversion functions.

fincs:
PR seems abandoned. Is there anyone willing to take it up and implement my last suggestion? (transparent UTF-8 <--> UTF-16 conversion)

Pirater12:
@fincs @joel16 If you want I can take over this PR

joel16:
Yeah go for it.

libctru/include/3ds/services/frd.h
```diff
@@ -141,7 +141,7 @@ Result FRD_IsFromFriendList(FriendKey *friendKeyList, bool *isFromList);
  * @brief Updates the game mode description string.
  * @param desc Pointer to write the game mode description to.
  */
-Result FRD_UpdateGameModeDescription(u16 *desc);
+Result FRD_UpdateGameModeDescription(u8 *desc);
```
fincs: Should be const char*

libctru/include/3ds/services/frd.h
```diff
@@ -119,7 +119,7 @@ Result FRD_GetMyFavoriteGame(u64 *titleId);
  * @brief Gets the current user's comment on their friend profile.
  * @param comment Pointer to write the current user's comment to.
  */
-Result FRD_GetMyComment(u16 *comment);
+Result FRD_GetMyComment(u8 *comment);
```
fincs: Should be char*, and should have a parameter to describe the output buffer size (and use it).

libctru/include/3ds/services/frd.h
```diff
@@ -93,9 +93,9 @@ Result FRD_GetMyProfile(Profile *profile);
 
 /**
  * @brief Gets the current user's screen name.
- * @param name Pointer to write the current user's screen name to. 11-byte UTF-16 screen name (with null terminator)
+ * @param name Pointer to write the current user's screen name to.
  */
-Result FRD_GetMyScreenName(u16 *name);
+Result FRD_GetMyScreenName(u8 *name);
```
fincs: Should be char*, and should have a parameter to describe the output buffer size (and use it).

libctru/include/3ds/services/frd.h
Pirater12: >  should have a parameter to describe the output buffer size
Hmmm, just strlen it?

fincs: I meant, the total capacity of the buffer the user provides to the function.

PR #406, opened 2018-05-28
Add apt function to allow disabling home button

astronautlevel2:
Add a function to allow enabling/disabling whether the home button is usable.

This allows homebrew to block the home button, similar to the eShop when loading, MH4U when online, and almost certainly other games/applications which I can't think of off the top of my head.

Kind of a hack and we're almost certainly forgetting something, but it seems to work fine.

### DIFF
--- a/libctru/source/services/apt.c
+++ b/libctru/source/services/apt.c
@@ -41,6 +41,7 @@ enum
 
 static u8 aptHomeButtonState;
 static u32 aptFlags = FLAG_ALLOWSLEEP;
+static bool home_allowed = false;
 static u32 aptParameters[0x1000/4];
 static u64 aptChainloadTid;
 static u8 aptChainloadMediatype;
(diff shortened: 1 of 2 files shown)


libctru/source/services/apt.c
```diff
@@ -41,6 +41,7 @@ enum
 
 static u8 aptHomeButtonState;
 static u32 aptFlags = FLAG_ALLOWSLEEP;
+static bool home_allowed = false;
```
fincs: Rename to aptHomeAllowed.

fincs:
In addition, there should probably be an API to detect whether HOME is pressed and it is blocked. Official apps display a cute interdictory "forbidden HOME" logo in the bottom screen when this happens.

astronautlevel2:
The function to check if HOME is blocked is already present - I can add the API to detect whether HOME is pressed tomorrow when I have a bit more time.

fincs:
Any updates on this PR?

fincs:
Pinging @astronautlevel2 again.

libctru/source/services/apt.c
```diff
@@ -41,6 +41,7 @@ enum
 
 static u8 aptHomeButtonState;
 static u32 aptFlags = FLAG_ALLOWSLEEP;
+static bool aptHomeAllowed = false;
```
Oreo639: I don't think it is a good idea to block going to the home menu by default.

Oreo639: I don't think it is a good idea to block the home menu by default.

fincs:
PR superseded by #430.

PR #428, opened 2019-04-01
Add Mii struct

hax0kartik:


libctru/include/3ds/mii.h
```diff
@@ -0,0 +1,123 @@
+#pragma once
+
+#include <3ds.h>
```
fincs: Don't include 3ds.h from libctru headers. Include `3ds/types.h`

libctru/include/3ds/services/frd.h
```diff
@@ -4,6 +4,7 @@
  */
 #pragma once
 #include <3ds.h>
```
fincs: I just realized that 3ds.h is also included from frd.h. Can you fix this so that the actual headers required by frd.h are included?

hax0kartik: Fixed + mcuhwc.c was having the same issue, fixed that too.

PR #430, opened 2019-04-24
Add apt function to allow disabling home button

Oreo639:
This pr replaces https://github.com/smealum/ctrulib/pull/406 as was discussed with asty and fincs.
Text below taken directly from said pr:

Add a function to allow enabling/disabling whether the home button is usable.

This allows homebrew to block the home button, similar to the eShop when loading, MH4U when online, and almost certainly other games/applications which I can't think of off the top of my head.

Kind of a hack and we're almost certainly forgetting something, but it seems to work fine.

### DIFF
--- a/libctru/source/services/apt.c
+++ b/libctru/source/services/apt.c
@@ -40,8 +40,9 @@ enum
 };
 
 static u8 aptHomeButtonState;
+static u8 aptRecentHomeButtonState;
 static u32 aptFlags = FLAG_ALLOWSLEEP;
-static bool aptHomeAllowed = false;
+static bool aptHomeAllowed = true;
 static u32 aptParameters[0x1000/4];
 static u64 aptChainloadTid;
 static u8 aptChainloadMediatype;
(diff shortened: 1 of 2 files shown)


libctru/source/services/apt.c
```diff
@@ -513,8 +532,15 @@ bool aptMainLoop(void)
 			GSPGPU_SetLcdForceBlack(0);
 		aptCallHook(APTHOOK_ONWAKEUP);
 	}
-	else if ((aptFlags & FLAG_POWERBUTTON) || aptHomeButtonState)
-		aptProcessJumpToMenu();
+	else if ((aptFlags & FLAG_POWERBUTTON) || aptHomeButtonState) {
```
fincs: Place `{` on new line.

libctru/source/services/apt.c
```diff
@@ -513,8 +532,15 @@ bool aptMainLoop(void)
...
+		if (aptHomeAllowed || (aptFlags & FLAG_POWERBUTTON))
+			aptProcessJumpToMenu();
+		else {
```
fincs: Same thing here (place `{` on new line).

PR #433, opened 2019-05-04
Implement Luma3DS gdb hio

TuxSH:
For debugged application, requires upcoming release/latest commit

### DIFF
--- a/libctru/source/gdbhio.c
+++ b/libctru/source/gdbhio.c
@@ -1,0 +1,397 @@
+#include <3ds/gdbhio.h>
+
+#include <3ds/svc.h>
+#include <3ds/result.h>
+
+#include <string.h>
+#include <errno.h>
+
+#include <stdarg.h>
+#include <stdio.h>
+
+#define GDBHIO_O_RDONLY           0x0
+#define GDBHIO_O_WRONLY           0x1
+#define GDBHIO_O_RDWR             0x2
+#define GDBHIO_O_APPEND           0x8
+#define GDBHIO_O_CREAT          0x200
+#define GDBHIO_O_TRUNC          0x400
+#define GDBHIO_O_EXCL           0x800
+#define GDBHIO_O_SUPPORTED	(GDBHIO_O_RDONLY | GDBHIO_O_WRONLY| \
+				 GDBHIO_O_RDWR   | GDBHIO_O_APPEND| \
+				 GDBHIO_O_CREAT  | GDBHIO_O_TRUNC| \
+				 GDBHIO_O_EXCL)
+
+#define GDBHIO_S_IFREG        0100000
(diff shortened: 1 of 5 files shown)


libctru/source/gdbhio.c
```diff
@@ -0,0 +1,397 @@
...
+{
+	return _gdbHioSendSyncRequest("open", "siI", pathname, _gdbHioExportOpenFlags(flags), _gdbHioExportFileMode(mode));
+}
+
+int gdbHioClose(int fd)
+{
+	return _gdbHioSendSyncRequest("close", "i", fd);
+}
+
+int gdbHioRead(int fd, void *buf, unsigned int count)
```
mtheall: Why not pass through `count` as `unsigned long` and use `L`?

libctru/source/gdbhio.c
```diff
@@ -0,0 +1,397 @@
...
+{
+	return _gdbHioSendSyncRequest("read", "ipI", fd, buf, count);
+}
+
+int gdbHioWrite(int fd, const void *buf, unsigned int count)
```
mtheall: Why not pass through `count` as `unsigned long` and use `L`?

libctru/source/gdbhio.c
```diff
@@ -0,0 +1,397 @@
...
+{
+	return _gdbHioSendSyncRequest("write", "ipI", fd, buf, count);
+}
+
+int gdbHioLseek(int fd, off_t offset, int flag)
```
mtheall: `off_t` is signed; does that mean you should be using `l`?

libctru/source/gdbhio.c
TuxSH: GDB uses unsigned int (32-bit) https://sourceware.org/gdb/onlinedocs/gdb/read.html#read

libctru/source/gdbhio.c
TuxSH: Same as above

libctru/source/gdbhio_dev.c
```diff
@@ -0,0 +1,196 @@
...
+
+int gdbHioDevGetStderr(void)
+{
+	return _gdbHioDevImportFd(GDBHIO_STDERR_FILENO);
+}
+
+int gdbHioDevRedirectStdStreams(bool in, bool out, bool err)
+{
+	int ret = 0;
+	if (in && (ret = dup2(gdbHioDevGetStdin(), STDIN_FILENO) < 0)) return ret;
```
WinterMute: The file descriptors returned from gdbHioDevGetStd* need closed here so the user can close the new descriptors & undo the redirection without leaking resources. dup2 increases the refcount on the cloned handle so it won't be released when the first descriptor that references it is closed.

PR #441, opened 2019-12-11
Make the FIONBIO ioctl actually work.

raxoft:
The fcntl needs the original `sockfd`, not the remapped one.


fincs:
Or better yet, remove that section. It will fail anyway if fnctl cannot find the socket fd.

raxoft:
Sure, your call. It may change the semantics a bit, reporting ENOTTY instead of invalid socket for other ioctls, but who cares, I guess...

PR #449, opened 2020-05-06
GPU HW init, etc.

TuxSH:
GSP assumes its first user will initialize the GPU (GSP only initializes the basic engines, leaving some blocks without clock enabled). This fixes it

See also: https://gist.github.com/TuxSH/40a861448be397347090975ac65a633b

### DIFF
--- a/libctru/include/3ds/services/gspgpu.h
+++ b/libctru/include/3ds/services/gspgpu.h
@@ -144,8 +144,11 @@ Result GSPGPU_ReleaseRight(void);
  */
 Result GSPGPU_ImportDisplayCaptureInfo(GSPGPU_CaptureInfo*captureinfo);
 
-/// Sames the VRAM sys area.
+/// Saves the VRAM sys area.
 Result GSPGPU_SaveVramSysArea(void);
+
+/// Resets the GPU
+Result GSPGPU_ResetGpuCore(void);
 
 /// Restores the VRAM sys area.
 Result GSPGPU_RestoreVramSysArea(void);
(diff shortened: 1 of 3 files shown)


profi200:
Huh? Is this really the case? I assumed GSP does the majority of these reg writes. It has code for almost all of that.

TuxSH:
> Huh? Is this really the case? I assumed GSP does the majority of these reg writes. It has code for almost all of that.

It certainly does, but only in `ResetGpuCore` or `AcquireRight(forced=true)` path

Otherwise it only touches `(PSC_ ?)CLK`, the reg @0x30 plus the PDC regs sets.

`PSC_CLK` is set to only 0x100 when gsp starts.

fincs:
Official software has a previously undiscovered codepath that performs these writes when GSPGPU_RegisterInterruptRelayQueue returns 0x2A07 (i.e. first time usage). This PR corrects libctru to perform this initial hardware setup; which allows the GPU to be used in homebrew that runs on boot prior to home menu (or mset) starting.

profi200:
Interesting. I saw inits happen multiple times in the Corgi3DS logs and thought this was GSP doing it.

And yeah, "PSC_CLK" is the official name. That reg actually has no visible effect when left to 0x100. But it's set to 0x70100 anyway. Only bit 0 has an effect. It seems to screw up every clock including display controller clocks (heavily corrupted graphics).

TuxSH:
> And yeah, "PSC_CLK" is the official name

I'm sure you have noticed the reg @ 0x30 (`SRAM_MODE`) then. I've confirmed that VRAM is controlled by this reg (1 bit per bank), meaning VRAM is split into 4 banks, not just 2; messing with this reg should be done only if the framebuffer isn't in vram

PR #450, opened 2020-05-09
Change allocateHeaps logic to properly handle appmemtype 5

TuxSH:


libctru/source/system/allocateHeaps.c
```diff
@@ -10,17 +10,25 @@ u32 __ctru_heap;
 u32 __ctru_linear_heap;
 
 __attribute__((weak)) u32 __ctru_heap_size        = 0;
-__attribute__((weak)) u32 __ctru_linear_heap_size = 32*1024*1024;
+__attribute__((weak)) u32 __ctru_linear_heap_size = 0;
 
 void __attribute__((weak)) __system_allocateHeaps(void) {
-	u32 tmp=0;
+	// By default, automatically allocate all remaining free memory, aligning to page size.
+	static const u32 max_linear_heap_size = 32 << 20;
```
fincs, line 17: Make this a #define

libctru/source/system/allocateHeaps.c
```diff
@@ -3,24 +3,33 @@
 #include <3ds/env.h>
 #include <3ds/os.h>
 
+#define MAX_LINEAR_HEAP_SIZE (32 << 20) // 32MB
```
fincs, line 6: Rename to `DEFAULT_LINEAR_HEAP_SIZE`, as this name might misleadingly imply that you can't have more than 32mb of linear heap no matter what (and you can override `__ctru_linear_heap_size` with a bigger size).

PR #451, opened 2020-05-09
Weaken more init/env functions

TuxSH:


libctru/source/system/initArgv.c
```diff
@@ -4,13 +4,13 @@
 #include <string.h>
 
 // System globals we define here
-int __system_argc;
-char** __system_argv;
+int __system_argc = 0;
```
fincs, line 7: Unneeded, as by default stuff goes into BSS.

PR #452, opened 2020-05-09
APT workaround when Home Menu is not launched, etc.

TuxSH:


libctru/source/services/apt.c
```diff
@@ -188,6 +189,8 @@ Result aptInit(void)
 
 	if (AtomicPostIncrement(&aptRefCount)) return 0;
 
+	aptHomeAllowed = osGetSystemCoreVersion() == 2;
```
fincs, line 192: Wouldn't it be a better idea to set a default for aptHomeAllowed using APT_IsRegistered(APPID_HOMEMENU, ...) just like in the chainload case?

TuxSH: No, because APT is allowed to launch Home Menu if it's not already been launched. For example if you open Pokémon Sun on O3DS, Home Menu isn't launched on boot.

PR #455, opened 2020-05-14
Add hidKeysDownRepeat

TuxSH:


libctru/source/services/hid.c
```diff
@@ -195,6 +204,21 @@ void hidScanInput(void)
 	kDown = (~kOld) & kHeld;
 	kUp = kOld & (~kHeld);
 
+	if (kDelay)
+	{
+		if (kDown)
```
mtheall, line 209: switch back to `if (kHeld != kOld)`

PR #458, opened 2020-05-20
Add mtime to stat

piepie62:
I would add this to fstat, but archive_getmtime requires a path. Not sure if there's a way around that.


mtheall:
This was deliberately left out because it is extremely slow, and also because there's no way to get the mtime without a path.

piepie62:
Ah. Fair enough, then; I didn't realize that.

PR #464, opened 2020-06-13
Update travis os for doxygen

Oreo639:
Xenial is too old to run the doxygen 1.8.18 release, this fixes that.
This also moves all the doxygen installing stuff into a script.

### DIFF
--- a/.travis/doxyprep.sh
+++ b/.travis/doxyprep.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+if [ ! -e "$DOXY_BINPATH/doxygen" ]; then
+	mkdir -p ~/doxygen && pushd ~/doxygen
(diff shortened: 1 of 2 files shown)


fincs:
Thank you for sparing me the misery of having to figure out why a Linux binary will fail to run on Linux during a library release.

.travis/doxyprep.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+if [ ! -e "$DOXY_BINPATH/doxygen" ]; then
+	mkdir -p ~/doxygen && pushd ~/doxygen
+	wget http://doxygen.nl/files/doxygen-1.8.18.linux.bin.tar.gz
```
profi200, line 5: This is already merged but i can't oversee the "http".

PR #470, opened 2020-07-14
Add LightEvent_WaitTimeout and remove magic numbers

Nanquitas:


fincs:
Suggestions:

- Separate the magic number changes into a different commit
- Rename to `LightEvent_WaitTimeout` to match `CondVar_WaitTimeout`
- `for (; res != timeoutRes; )` -> `while (res != timeoutRes)`

Issue #404, opened 2018-05-19
CSND adpcm state address hardcoded


mariohackandglitch:
https://github.com/smealum/ctrulib/blob/fcf9ca767b640965b9e98c482d3e0cfaca862a59/libctru/source/services/csnd.c#L530-L535

This part of the code is hardcoded to get the adpcm state from the position before the data buffer. While this may be "useful" in some case, in others where the data is being read from a file (bcwav in this case) it is not useful at all. If it wasn't hardcoded it would be possible to just load the file to memory and process the data there, however since the data buffers for different channels are next to each other, it's not possible to use that approach.

Note that while it's not recommended to use csnd to play music, it is very useful for sound effects.

fincs:
CSND should be used anymore, and we don't plan adding any features to it at all.

There's no reason to choose CSND over NDSP for the purposes of playing sound effects. NDSP gives you 24 independently mixed sound channels (just like CSND), and way more control over what is played and effects.

mariohackandglitch:
It's the only workaround if ndsp is not available.

In any case, why have something that partly works and not just remove it at all?

profi200:
It has its use cases. It's the only one which would reliably work from a different (background) process alongside a game/homebrew.

mariohackandglitch:
What profi200 said, that's why I want to use it. DSP service is being used by the game and it's not a good idea to use it. csnd in the other hand works flawlessly.

PR #478, opened 2020-12-20
CSND improvements.

mariohackandglitch:
This pull request adds two separate improvements to the csnd service implementation.

### Commit 0731ba8
As previously mentioned in issue #404, the current implementation hardcodes the ADPCM state variables to be placed before the first sound data buffer. This makes the usage of the service less practical for CWAV playback, as those variables are stored in the INFO header block, so they have to be set manually. (Or copied before the data block, which may overwrite other CWAV stuff.)

This commit adds `csndSetNextAdpcmState`, which sets the ADPCM state for the next `csndPlaySound` call (after which it is cleared). The implementation was made in a way that won't break any previous homebrew that uses csnd to play ADPCM sound data.

### Commit 57764a3
In the current implementation, the sound data buffers are limited to the 0x14000000 memory region, which may not be available in some situations (such as .3GX game plugins).

This commit splits `csndPlaySound` into `csndPlaySound` and `csndPlaySoundPhys`. The first function has the same behaviour as the current implementation, while the second one takes physical addresses for the sound data buffers. This way, the user is the responsable for converting the virtual address into the proper physical address (for example, using Luma3DS custom SVC). The split has been made in order to keep compatibility with any previous homebrew (instead of adding a flag to indicate if the address is virtual or physical).

### Commit 4d892fa, 3e84363
This commit adds and implements `CSND_PlaySoundDirectly`, as well as the interfaces `csndInitializeDirectSound` and `csndPlayDirectSound`. It turns out the memory at csndSharedMem + offset[3] holds the information for playing a "direct sound". I recreated this info with the `CSND_DirectSound` struct. However, there are a few values I don't yet understand, so it may be wise to wait for someone who has researched this in case they have more information.
Implementation example: https://github.com/mariohackandglitch/libcwav/blob/main/source/cwav.c#L301-L363 and https://github.com/mariohackandglitch/libcwav/blob/main/example_libcwav/source/main.cpp#L149-L165

## Why?
In issue #404, it was mentioned that "csnd should not be used anymore" and "there is no reason to choose CSND over NDSP", however, this is not true for the following reasons:

1. CSND seems to be used mainly for .bcwav playback. This can be very easily deduced by looking at the bcwav file structure and the csnd available service calls and commands. (Of course, there is no way to confirm this, so this reason doesn't have much weight if any, as DSP can be used as well.)
2. **DSP service may not always be available!**. At the time of writing, the only situation where this occurs is with 3gx game plugins, as they run on top of already running applications which have their own DSP session and communication channel (making NDSP unusable). In fact, this is probably why (afaik) applets use the CSND service instead of DSP, so they don't interfere with suspended applications which may already be using DSP.

Moreover, the changes in this pull request were made with compatibility in mind, so they won't affect any existing homebrew applications.
That said, there is absolutely no reason (that I can think of, please prove me wrong) for keeping csnd in the current state and not applying the changes here, which improve the legit use cases of the service.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Version 2.0.1
+
+- Added CFG_SystemModel enum.
+- Fixed bug in condvar code.
+- Fixed bug in srvpm code.
+- Fixed const correctness issues in gspgpu code.
+- Further improvements to overall system stability and other minor adjustments have been made to enhance the user experience.
+
 ## Version 2.0.0
 
 **This is a major release. Many essential components have been overhauled, although breaking changes are minimal and only affect rarely used functionality.**
--- a/libctru/Makefile
+++ b/libctru/Makefile
@@ -10,7 +10,7 @@ include $(DEVKITARM)/base_rules
 
 export LIBCTRU_MAJOR	:= 2
 export LIBCTRU_MINOR	:= 0
-export LIBCTRU_PATCH	:= 0
+export LIBCTRU_PATCH	:= 1
 
 
 VERSION	:=	$(LIBCTRU_MAJOR).$(LIBCTRU_MINOR).$(LIBCTRU_PATCH)


fincs:
A few comments:

- This PR has been submitted a day after we're done releasing a new version of libctru, which is quite unfortunate. We'd recommend developers intending to work on non-trivial PRs to give us a heads up first so that we can coordinate better.
- ADPCM state is handled the way it is right now in order to mimic how the old sound hardware in the Nintendo DS did it. I agree in retrospect this might not have been exactly the brightest idea, so a minimally invasive solution that solves this issue is appreciated.
- osConvertVirtToPhys supports converting addresses in other linearly mapped memory regions other than the legacy 0x14xxxxxx FCRAM memory region - please check out its implementation. Assuming you meant allowing usage of non-linearly mapped regions, I'm not sure when or how would that be required. Using non-device memory with devices feels like a contradiction, and if your use case requires it I'd advise to instead figure out how to properly create a block of device memory. Keep in mind that the OS makes no guarantees whatsoever about the contiguity of non-device memory, so in effect you are currently relying on undefined behavior.
- The main use case of CSND as shown in 3DS Horizon is providing applets with the ability to play sound on top of games (which use the DSP). It is unrelated to bcwav (which is just a container format part of Nintendo's official middleware tools). Currently the APT logic in libctru is not prepared to handle existing in any slot other than the Application slot; that's a bigger problem that needs to be solved first before we even get to the situation in which using CSND makes sense.
- I don't know what "3gx game plugins" are, but I assume they are some form of injected code that runs inside game processes. This has never been a supported use case of libctru (and will never be), however I'm willing to consider contributions that are generally beneficial.

mariohackandglitch:
Thanks for reading the pull request, I'll elaborate on your comments:
- Sorry about this, I was cleaning up some code I had written in the past and found the csnd file that I had to override due to the problem described in the OP. This happened only a few hours ago, so I decided to make the actual implementation of the solution myself this time (as opposed to #404 where I only reported the issue) and didn't notice a release had just came out. I'll keep it in mind for next time.
- I'm not sure if you are affiriming that my solution is minimally invasive, or that it should be less invasive. I proposed this solution because I couldn't think of another way to achieve it without breaking compatibility with existing homebrew application source code that may still use csnd.
- (comments 3, 4 and 5): As you stated, 3gx game plugins consist of an executable (text and data) that is injected into game processes to expand their functionality. Obviously, the memory mapped for the executable cannot overlap with the game, so a kernel extension is used to map the plugin related linear memory (executable and heap) to unused regions (0x07000000 and 0x06000000 respectively). When plugins were first developed, they had to include a stripped down version of libctru. However, as time passed, ways were found to use official libctru service implementations while only overriding the needed parts. (In fact, commit 58c1bae was made with the purpose of making this process easier). 3GX plugin code makes use of a notification system exposed by Rosalina to recieve notifications that APT would normally handle, as the host game is already using an APT session for that.

The only problem we have faced with this approach is the one reported here. We are lucky that the 3ds offers a second sound service that we can use for playing sound on top of the game, instead of hacking into the game DSP implementation (like we had to do with GSP to display stuff to the screen). Every other service (with some limitations) works as fine as they do without any libctru modifications.

fincs:
> I'm not sure if you are affiriming that my solution is minimally invasive, or that it should be less invasive

I meant that it *is* minimally invasive, heh.

>  Obviously, the memory mapped for the executable cannot overlap with the game, so a kernel extension is used to map the plugin related linear memory (executable and heap) to unused regions (0x07000000 and 0x06000000 respectively).

For what it's worth, executable/regular heap is not supposed to be sourced from device memory - only memory that will be used with devices should be sourced from device memory. In addition, having the device memory be mapped at the correct VA shouldn't interfere with the game's operations, since the game and the plugin would be occupying different areas of VA space within the same FCRAM linear mapping region anyway. I guess mapping code/non-device-data to a custom VA region is somewhat reasonable, but mapping device memory outside the device memory region is not and I don't see any reason why it would be necessary.

I'm not sure how I feel about the described plugin system. libctru is certainly not built or designed around the "coexisting with other code, especially official" use case. Any attempt at doing this is bound to be unsupported, easy to break and difficult to maintain :\

mariohackandglitch:
> For what it's worth, executable/regular heap is not supposed to be sourced from device memory - only memory that will be used with devices should be sourced from device memory. In addition, having the device memory be mapped at the correct VA shouldn't interfere with the game's operations, since the game and the plugin would be occupying different areas of VA space within the same FCRAM linear mapping region anyway. I guess mapping code/non-device-data to a custom VA region is somewhat reasonable, but mapping device memory outside the device memory region is not and I don't see any reason why it would be necessary.

I'm getting a bit lost here, sorry. The master mind behind the core plugin system is @Nanquitas, I've only been responsible on collaborating with the maintenance of the existing plugin loader code, so I don't have a full understanding on every single quirk it has. :P
Here is the relevant part of the code about memory mapping: https://github.com/Nanquitas/Luma3DS/blob/3gx-loader/sysmodules/rosalina/source/plugin/memoryblock.c#L13-L90
Here is svcControlMemoryUnsafe: https://github.com/Nanquitas/Luma3DS/blob/3gx-loader/k11_extension/source/svc/ControlMemoryUnsafe.c

> I'm not sure how I feel about the described plugin system. libctru is certainly not built or designed around the "coexisting with other code, especially official" use case. Any attempt at doing this is bound to be unsupported, easy to break and difficult to maintain :\
We won't be making a full reimplementation of the communications with system services while we can use libctru existing implementation. This makes plugins difficult to mantain when something important in libctru changes (has happened before), but there's nothing else we can do about it. 

In any case, I'm not sure how this all affects csnd. It's an available system service like all of the others (even if it's not normally used), so I don't see the reason to remove it's implementation. Another solution would be to provide csnd as a standalone library, which would actually make sense (but would take a bit of work to implement) .

mariohackandglitch:
Relevant to this pull request:
I've been researching an applet csnd implementation, and it looks like the code responsible for parsing the bcwav format is called directly from the csnd api, so this reinforces the theory that csnd is built around bcwav playback.
Also looks like applets use the undocumented ExecuteType1Commands (0x00040080) command (by writing stuff directly to the csnd shared memory (?)).

mariohackandglitch:
More research on the undocumented command 0x00040080 (named it CSND:PlaySoundDirectly):

### Commit 4d892fa
This commit adds and implements `CSND_PlaySoundDirectly`, as well as the interfaces `csndInitializeDirectSound` and `csndPlayDirectSound`. It turns out the memory at csndSharedMem + offset[3] holds the information for playing a "direct sound". I recreated this info with the `CSND_DirectSound` struct. However, there are a few values I don't yet understand, so it may be wise to wait for someone who has researched this in case they have more information.
Implementation example: https://github.com/mariohackandglitch/libcwav/blob/main/source/cwav.c#L301-L363 and https://github.com/mariohackandglitch/libcwav/blob/main/example_libcwav/source/main.cpp#L149-L165

With this commit, a new use case can be given to CSND, as it's how the system plays sounds ignoring the sound slider and headphones (for example, the camera shutter sound).

mariohackandglitch:
Commit 3e84363 adds the latest research to `CSND_PlaySoundDirectly`. The other unknown parameters in the `CSND_DirectSound` doesn't seem to affect sound playback (at least that I noticed). So I think that for now it would be a good idea to implement the changes, seeing that nobody has the info about the remaining parameters, which could be added in the future.

mariohackandglitch:
Just leaving here another use case for CSND:
https://gbatemp.net/threads/poc-developer-only-background-3ds-mp3-player.589949/

mariohackandglitch:
With devkitPro's failure to communicate with developers, I don't really feel like contributing to your projects anymore.
The research made in this PR will be put into a standalone library so users will be able to use it.

WinterMute:
People are busy and don't often have the spare cycles to deal with some PRs as quickly as the submitters may like. Being patient and not linking to sites that promote copyright infringement would be helpful. Deleting PRs and the repositories behind them seems unecessarily petty tbh.

mariohackandglitch:
Finally I can get a reply back from you after several months! I don't really get how else I should react if a comment that tries to contribute to a discussion gets deleted with absolutely 0 context. I remind you that there is a feature to edit other's comments, and snip the link from it in a more polite way. The PR was closed for the reason described in my comment and the repo so I'm able to have the "forked from" status in another repository under my account.

In any case, I apologize for the sudden closing of the PR, but I still think communication with developers should be improved.

We can give this PR another try if you would like to.

PR #495, opened 2022-05-28
Implement `clock_gettime` syscall

rHermes:
This implements the syscalls for `clock_gettime` and `clock_getres`. We
support two clocks: CLOCK_REALTIME and CLOCK_MONOTONIC. I've opted to
use the existing `osGetTime()` code for the realtime clock, because it's
known to work.

For CLOCK_MONOTONIC I've used `svcGetSystemTick()` directly, as it has a
higher resolution. We can ignore the drift and so on, because it's
supposed to be just the number of ticks since last boot.

Implementing the `clock_gettime` syscall allows `std::chrono` to work, which is the primary reason I started looking at this.

There are two points that I am a bit unsure about:
- Is it ok to cap the precision of the monotonic clock to 1 microsecond? There are about 268 ticks per microsecond, so we could potentially lower the precision more. I don't know how important this is, but it could be a potential increase.
- Should I be using the `osGetTimeRef()` function and `osTimeRef_s.value_tick` in the monotonic timer? As far as I understand this is just an offset as to when a reference point was set, and is not needed here, but I could be wrong here.

Last, I want to thank you for creating such a library so awesome that even beginner like me can explore the 3ds!

### DIFF
--- a/libctru/source/system/syscalls.c
+++ b/libctru/source/system/syscalls.c
@@ -3,12 +3,15 @@
 #include <sys/lock.h>
 #include <sys/reent.h>
 #include <string.h>
+#include <time.h>
+#include <errno.h>
 
 #include <3ds/types.h>
 #include <3ds/svc.h>
 #include <3ds/env.h>
 #include <3ds/os.h>
 #include <3ds/synchronization.h>
+
 #include "../internal.h"
 
 void __ctru_exit(int rc);
@@ -28,6 +31,64 @@ struct _reent* __SYSCALL(getreent)()
 	return tv->reent;
 }
 
+int __SYSCALL(clock_gettime)(clockid_t clock_id, struct timespec *tp) {
+
+	if (clock_id == CLOCK_REALTIME)
+	{
+		if (tp != NULL)
+		{
+			// Retrieve current time, adjusting epoch from 1900 to 1970
+			s64 ms_since_epoch = osGetTime() - 2208988800000ULL;
+			tp->tv_sec = ms_since_epoch / 1000;
+			tp->tv_nsec = (ms_since_epoch % 1000) * 1000000;
+		}
+	}
+	else if (clock_id == CLOCK_MONOTONIC) 
+	{
+		if (tp != NULL)
+		{
+			// Use the ticks directly, as it offer the highest precision
+			u64 ticks_since_boot = svcGetSystemTick();
+
+			tp->tv_sec = ticks_since_boot / SYSCLOCK_ARM11;
+			tp->tv_nsec = ((ticks_since_boot % SYSCLOCK_ARM11) * 1000000000ULL) / SYSCLOCK_ARM11;
+		}
+	}
+	else
+	{
+		errno = EINVAL;
+		return -1;
+	}
+
+	return 0;
+}
+
+int __SYSCALL(clock_getres)(clockid_t clock_id, struct timespec *res) {
+	if (clock_id == CLOCK_REALTIME)
+	{
+		if (res != NULL) 
+		{
+			res->tv_sec = 0;
+			res->tv_nsec = 1000000;
+		}
+	}
+	else if (clock_id == CLOCK_MONOTONIC)
+	{
+		if (res != NULL)
+		{
+			res->tv_sec = 0;
+			res->tv_nsec = 1;
+		}
+	}
+	else
+	{
+		errno = EINVAL;
+		return -1;
+	}
+
+	return 0;
+}
+
 //---------------------------------------------------------------------------------
 int __SYSCALL(gettod_r)(struct _reent *ptr, struct timeval *tp, struct timezone *tz) {
 //---------------------------------------------------------------------------------
@@ -37,7 +98,7 @@ int __SYSCALL(gettod_r)(struct _reent *ptr, struct timeval *tp, struct timezone 
 
                 // Convert to struct timeval
                 tp->tv_sec = now / 1000;
-                tp->tv_usec = (now - 1000*tp->tv_sec) * 1000;
+                tp->tv_usec = (now % 1000) * 1000;
         }
 
         if (tz != NULL) {


mtheall:
Making your own std::chrono clock is fairly trivial:
```cpp
#include <3ds.h>

#include <chrono>
#include <cinttypes>
#include <cstdint>
#include <cstdio>
#include <cstdlib>
#include <cstring>
#include <ctime>
#include <ratio>

class system_clock
{
public:
	using rep        = std::uint64_t;
	using period     = std::milli;
	using duration   = std::chrono::duration<rep, period>;
	using time_point = std::chrono::time_point<system_clock>;

	static constexpr bool is_steady = false;

	static std::time_t to_time_t(time_point const &t) noexcept
	{
		return std::time_t(std::chrono::duration_cast<std::chrono::seconds>(now().time_since_epoch()).count());
	}

	static time_point from_time_t(std::time_t t) noexcept
	{
		return time_point(duration(t));
	}

	static time_point now() noexcept
	{
		// convert 1900-01-01 epoch to unix epoch
		return time_point(duration(osGetTime() - 2208988800000ULL));
	}
};

class high_resolution_clock
{
public:
	using rep        = std::uint64_t;
	using period     = std::ratio<1, SYSCLOCK_ARM11>;
	using duration   = std::chrono::duration<rep, period>;
	using time_point = std::chrono::time_point<high_resolution_clock>;

	static constexpr bool is_steady = true;

	static time_point now() noexcept
	{
		return time_point(duration(svcGetSystemTick()));
	};
};

using steady_clock = high_resolution_clock;

int main(int argc, char* argv[])
{
	auto const start = high_resolution_clock::now();

	gfxInitDefault();
	consoleInit(GFX_TOP, NULL);

	while (aptMainLoop())
	{
		gspWaitForVBlank();
		gfxSwapBuffers();
		hidScanInput();

		u32 kDown = hidKeysDown();
		if (kDown & KEY_START)
			break;

		if (kDown & KEY_A)
		{
			auto const now = high_resolution_clock::now();
			std::printf("Tick %" PRIu64 ", %lf seconds since start\n",
				now.time_since_epoch().count(),
				std::chrono::duration<double>(now - start).count());
		}

		if (kDown & KEY_B)
		{
			auto const now = system_clock::to_time_t(system_clock::now());
			std::printf("%s", std::ctime(&now));
		}
	}

	gfxExit();
	return 0;
}
```

rHermes:
Thanks for the tip on using a custom implementation, but that seems to me to be the wrong route to take in this case. If we are not interested in implementing `clock_gettime`, then we should disable the use of it when compiling GCC. There are no warnings as of now that the standard behavior of the C++ library doesn't work.

Your use of the `period` for the steady_clock is interesting though, and we might be interested in putting those two implementations directly into the GCC directory, but the direct link to `libctru` would be problematic.

Overall, I think it's best to just implement the syscalls, as it's the most portable and solves the chrono issue in the same instance.

libctru/source/system/syscalls.c
```diff
@@ -28,6 +31,64 @@ struct _reent* __SYSCALL(getreent)()
...
+int __SYSCALL(clock_gettime)(clockid_t clock_id, struct timespec *tp) {
+
+	if (clock_id == CLOCK_REALTIME)
+	{
+		if (tp != NULL)
+		{
+			// Retrieve current time, adjusting epoch from 1900 to 1970
+			s64 ms_since_epoch = osGetTime() - 2208988800000ULL;
+			tp->tv_sec = ms_since_epoch / 1000;
+			tp->tv_nsec = (ms_since_epoch - 1000*tp->tv_sec) * 1000000;
```
mtheall, line 43: I used
```cpp
tp->tv_nsec = (ms_since_epoch % 1000) * 1000000;
```
and it ended up with less code, presumably because it can use the remainder output from the previous division operation.

libctru/source/system/syscalls.c
```diff
@@ -28,6 +31,64 @@ struct _reent* __SYSCALL(getreent)()
...
+	}
+	else if (clock_id == CLOCK_MONOTONIC) 
+	{
+		if (tp != NULL)
+		{
+			// Use the ticks directly, as it offer the highest precision
+			u64 usec_since_boot = svcGetSystemTick() / CPU_TICKS_PER_USEC;
+
+			tp->tv_sec = usec_since_boot / 1000000;
+			tp->tv_nsec = (usec_since_boot - 1000000*tp->tv_sec) * 1000;
```
mtheall, line 54: My version:
```cpp
            u64 ticks_since_boot = svcGetSystemTick();

            tp->tv_sec  = ticks_since_boot / SYSCLOCK_ARM11;
            tp->tv_nsec = (ticks_since_boot % SYSCLOCK_ARM11) * 1000000000ULL / SYSCLOCK_ARM11;
```
This gives about 3.7ns resolution.

I don't know why, but your version does some floating-point conversions but mine didn't.

libctru/source/system/syscalls.c
```diff
@@ -28,6 +31,64 @@ struct _reent* __SYSCALL(getreent)()
...
+			res->tv_sec = 0;
+			res->tv_nsec = 1000000;
+		}
+	}
+	else if (clock_id == CLOCK_MONOTONIC)
+	{
+		if (res != NULL)
+		{
+			res->tv_sec = 0;
+			res->tv_nsec = 1000;
```
mtheall, line 80: With the above change:
```cpp
res->tv_nsec = 1;
```

mtheall (commented):
Unfortunately GCC aliases high_resolution_clock to system_clock. If you actually want a high-resolution clock on 3DS with this code then use steady_clock.

libctru/source/system/syscalls.c
rHermes: Yeah, I should have seen that, I just copied what was done in `gettod_r`. I'll change it in both places.

libctru/source/system/syscalls.c
rHermes: Thanks, that's much better!

libctru/source/system/syscalls.c
rHermes: The real resolution is a bit lower than that, but since it's not a whole number, I guess 1 is the best we can do?

mtheall: Yeah I just used 1 since we're on that order of magnitude and also because we could get every possible value of tv_nsec.

rHermes:
Any update on this?

PR #501, opened 2022-07-18
Replaced the typedef of Thread in order to allow user redefinition

SeleDreams:
a pull request to allow easy implementation of my fix suggested in https://github.com/devkitPro/libctru/issues/500


fincs:
See https://github.com/devkitPro/libctru/issues/500#issuecomment-1193095215 .

PR #503, opened 2022-08-10
Include for size_t

csnikki:
For standalone usage.


fincs:
I'm merging this, out of correctness. However, programs should generally not rely on piecemeal libctru header includes, as we make no guarantee they will continue to exist under the same names and/or contents. `#include <3ds.h>` should be the canonical way to use libctru in a program.

PR #504, opened 2022-08-18
Use __tdata_align to align thread local storage

### LINKED ISSUE #497: TLS memory errors with aligned thread-local variables
Born from: https://github.com/Meziu/ctru-rs/issues/60
## Description
After some testing on our Rust libraries for this system, we've noticed a weird bug with the use of thread-local storage. With some investigation, we have pinpointed the problem to the use of *aligned* thread-local variables. As our tests show, the offset of the variable pointers are wrong when one or more aligned TLS variables exist, causing memory leaks, access of uninitialized memory etc. This issue appears in both Rust and C programs.

Furthermore, this issue seems to only affect (at least in our examples) programs with a `debug` optimization level.

E.g. :
`debug` mode in Rust reproduces the bug.
`release` mode in Rust doesn't reproduce the bug.
`-Og` in C reproduces the bug.
`-O2` in C doesn't reproduce the bug.

## Here is a test program to try the issue. Remember to use `-Og` to replicate:
```C
#include <3ds.h>
#include <stdio.h>
#include <string.h>

typedef ALIGN(4) struct {
    u8 inner[3];
} Align4;

typedef ALIGN(16) struct {
    u8 inner[3];
} Align16;

static __thread Align4 BUF_4 = {.inner = {2, 2, 2}};
static __thread Align16 BUF_16 = {.inner = {1, 1, 1}};

int
main(int argc, char** argv)
{
    gfxInitDefault();
    consoleInit(GFX_TOP, NULL);

    BUF_16.inner[0] = 0;

    bool reproduced = false;

    // Check if at least one byte has been offset improperly
    printf("[");
    for (int i = 0; i < 3; i++) {
        if (BUF_4.inner[i] != 2) {
            reproduced = true;
        }
        printf("%d, ", BUF_4.inner[i]);
    }
    printf("]\n");

    if (reproduced) {
        printf("Bug has been reproduced!\n");
    }
    else {
        printf("There has been no issue!");
    }

    // Main loop
    while (aptMainLoop()) {
        gspWaitForVBlank();
        hidScanInput();

        u32 kDown = hidKeysDown();
        if (kDown & KEY_START)
            break;  // break in order to return to hbmenu

        // Flush and swap framebuffers
        gfxFlushBuffers();
        gfxSwapBuffers();
    }

    gfxExit();
    return 0;
}
```

## Notes about our Rust toolchain
Our Rust programs mainly use `libctru` (to substitute missing `newlib` functions), and the `arm-none-eabi-gcc` linker. Most of the linking flags are the same (basically those in a normal Makefile but in the [Rust target-specification](https://github.com/rust-lang/rust/blob/master/compiler/rustc_target/src/spec/armv6k_nintendo_3ds.rs)). Either way, it's hard to believe this one to be an issue in either of the compilers (`rustc` and `gcc`) as the issue reproduces in both. It's likely an issue with `libctru` or the linker.

ian-h-chamberlain:
Fixes #497 

Accompanying PR: https://github.com/devkitPro/devkitarm-crtls/pull/6

Cases I was able to test:
Create `static __thread` variables of the following types and verify their values are initialized as expected, both in the main thread and one created with `threadCreate`:

* Align of 4 then align of 16
* Align of 4 then u8
* Align 16 then align of 4
* Align of 16 then u8
* u8 then align of 4
* u8 then align of 16
* u8 then u16
* u16 then u8

### DIFF
--- a/libctru/source/internal.h
+++ b/libctru/source/internal.h
@@ -8,6 +8,12 @@
 #define THREADVARS_MAGIC  0x21545624 // !TV$
 #define FS_OVERRIDE_MAGIC 0x21465324 // !FS$
 
+extern const size_t __tdata_align;
+extern const u8 __tdata_lma[];
+extern const u8 __tdata_lma_end[];
+extern u8 __tls_start[];
+extern u8 __tls_end[];
+
 // Keep this structure under 0x80 bytes
 typedef struct
 {
@@ -48,3 +54,11 @@ static inline ThreadVars* getThreadVars(void)
 }
 
 void initThreadVars(struct Thread_tag *thread);
+
+static inline size_t getThreadLocalStartOffset(size_t tls_tp) {
+	size_t align = 8;
+	if (__tdata_align > align) align = __tdata_align;
+	// ARM ELF TLS ABI mandates an 8-byte header, so we include an extra 8 bytes
+	// then add padding to align the .tdata properly
+	return (8 + (size_t)tls_tp + (__tdata_align - 1)) & ~(__tdata_align - 1);
+}
--- a/libctru/source/system/syscalls.c
+++ b/libctru/source/system/syscalls.c
@@ -16,10 +16,6 @@
 
 void __ctru_exit(int rc);
 
-extern const u8 __tdata_lma[];
-extern const u8 __tdata_lma_end[];
-extern u8 __tls_start[];
-
 struct _reent* __SYSCALL(getreent)()
 {
 	ThreadVars* tv = getThreadVars();
@@ -43,7 +39,7 @@ int __SYSCALL(clock_gettime)(clockid_t clock_id, struct timespec *tp) {
 			tp->tv_nsec = (ms_since_epoch % 1000) * 1000000;
 		}
 	}
-	else if (clock_id == CLOCK_MONOTONIC) 
+	else if (clock_id == CLOCK_MONOTONIC)
 	{
 		if (tp != NULL)
 		{
@@ -66,7 +62,7 @@ int __SYSCALL(clock_gettime)(clockid_t clock_id, struct timespec *tp) {
 int __SYSCALL(clock_getres)(clockid_t clock_id, struct timespec *res) {
 	if (clock_id == CLOCK_REALTIME)
 	{
-		if (res != NULL) 
+		if (res != NULL)
 		{
 			res->tv_sec = 0;
 			res->tv_nsec = 1000000;
@@ -168,7 +164,7 @@ void initThreadVars(struct Thread_tag *thread)
 	tv->thread_ptr = thread;
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Warray-bounds"
-	tv->tls_tp = (thread != NULL ? (u8*)thread->stacktop : __tls_start) - 8; // Arm ELF TLS ABI mandates an 8-byte header
+	tv->tls_tp = (thread != NULL ? (u8*)thread->stacktop : __tls_start);
 #pragma GCC diagnostic pop
 	tv->srv_blocking_policy = false;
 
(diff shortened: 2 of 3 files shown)


libctru/source/system/syscalls.c
```diff
@@ -168,7 +164,7 @@ void initThreadVars(struct Thread_tag *thread)
 	tv->thread_ptr = thread;
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Warray-bounds"
-	tv->tls_tp = (thread != NULL ? (u8*)thread->stacktop : __tls_start) - 8; // Arm ELF TLS ABI mandates an 8-byte header
+	tv->tls_tp = (thread != NULL ? (u8*)thread->stacktop : __tls_start);
```
ian-h-chamberlain, line 167: This is technically a change in behavior. 8 additional bytes are allocated for each thread to be used as the header, which seemed necessary for the alignment to be correct as I was experimenting, but there might be a way to avoid it.

libctru/source/internal.h
```diff
@@ -8,6 +8,12 @@
 #define THREADVARS_MAGIC  0x21545624 // !TV$
 #define FS_OVERRIDE_MAGIC 0x21465324 // !FS$
 
+extern const size_t __tdata_align;
```
ian-h-chamberlain, line 11: I think this could be a weak symbol if backwards compatibility with older versions of `devkitarm-crtls` is needed, which could fall back to the old behavior (align of 8).

libctru/source/system/syscalls.c
fincs: The double pointer sized TLS header is only used by more general, shared-library-oriented TLS models that we don't use (we effectively use `local-exec`), thus the reason why libctru (and libnx) have always faked the existence of the header by simply subtracting its size from the real start of the TLS area. I am not sure why you have had problems with alignment, I can only think of the fact that the *stack* on 32-bit machines typically needs 8-byte alignment, so that should be the baseline.

libctru/source/internal.h
fincs: Backwards compatibility is not a concern. libctru and devkitarm-crtls will be updated in lockstep.

libctru/source/system/syscalls.c
ian-h-chamberlain: It took a bit of trial and error to get right, since the location of the header also needs to be aligned properly, but I think I figured it out, and pushed my changes.

This appears to work with all the test cases I listed above in the description as well.

gyrovorbis:
Hey, guys, so this is going to sound kind of random, buuuut...

I'm a developer on the Sega Dreamcast indie SDK, KallistiOS, and I recently finished up my own quest to implement static TLS on the thing, and followed in the footsteps you guys laid out here and some of the alignment issues you ran into. Wound up making a really big multithreaded test suite just doing terrible things with .tdata and .tbss manually, automatically, and over-aligned variables to get things working.

First of all, thanks. Secondly, I just wanted to make sure I'm not missing something and you guys don't have any issues, but don't you also have to align the .tbss data similarly? Check what I had to do here:
https://github.com/KallistiOS/KallistiOS/pull/111

Anyway, kudos from the DC scene. Looking forward to working with DevkitPro some day with my own engine. :)

ian-h-chamberlain:
> First of all, thanks. Secondly, I just wanted to make sure I'm not missing something and you guys don't have any issues, but don't you also have to align the .tbss data similarly? Check what I had to do here:

@gyrovorbis wow, I'm amazed that this somehow helped someone on another project! Glad you were able to find it and found it useful for your purposes as well.

As for the alignment of `.tbss`, we did find two issues which led to https://github.com/devkitPro/devkitarm-crtls/pull/7:
1. Our linker script was forcing alignment incorrectly for both `.tdata` and `.tbss` sections – `ALIGN(4) :` vs `: ALIGN(4)` 🤦
2. Even after fixing that, we weren't accounting for the proper alignment of the `.tbss` section before marking the `__tls_end`, which resulted in the wrong size for the main thread's local storage section. Part of the PR above was also adding some extra padding using `ALIGNOF(.tbss)`, which as far as I could tell was sufficient to resolve the problem.

It looks like your code is doing some of the work in its `thd_create_tls_data` function instead of the linker script, but I think the end result is about the same? You might have a better understanding, having encountered this problem more recently and being more familiar with the KallistiOS code — if you think we're still missing something it might be good to revisit on our side too. Thanks for reaching out!

PR #505, opened 2022-08-28
add various ndspGet* methods

TurtleP:
This pull request add various `ndspGet*` methods for convenience *and* because they are tracked internally by libctru, applications cannot get this data otherwise. Most notably missing was `ndspGetMasterVol`, but I decided to add the remaining ones for consistency.

### DIFF
--- a/libctru/include/3ds/ndsp/ndsp.h
+++ b/libctru/include/3ds/ndsp/ndsp.h
@@ -119,10 +119,22 @@ u32    ndspGetFrameCount(void);
 void ndspSetMasterVol(float volume);
 
 /**
+ * @brief Gets the master volume.
+ * @return The master volume.
+ */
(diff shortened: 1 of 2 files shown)


libctru/include/3ds/ndsp/ndsp.h
```diff
@@ -118,24 +118,48 @@ u32    ndspGetFrameCount(void);
  */
 void ndspSetMasterVol(float volume);
 
+/**
+ * @brief Gets the master volume.
+ * @return The master volume.
+ */
+void ndspGetMasterVol(void);
```
fincs, line 125: void return type? :p

TurtleP: woops, let me fix that one haha

PR #510, opened 2022-09-20
Enable offline documentation by providing docsets

dimaguy:
Hello, I've gotten very frustrated at doxygen's search feature not showing me the suitable functions for my objectives right away (be it because the name of the function doesn't start *exactly* by the search string or by being slow at doing it's job, it's laggy).  
So I've found an article in hackernews promoting the wonders of offline documentation and showed docsets as a simple but effective option at doing so(while they were created for apple computers, with the right apps, like zeal which is free and cross platform, one can make full use of them in any computer). I did some research and came up with the knowledge needed to make doxygen generate the raw docsets and a way to index them and make them proper usable
This PR request sets the workflow to provide the docsets in artifacts, but they could probably be moved somehow to releases.

### DIFF
--- a/.github/workflows/doxygen.yaml
+++ b/.github/workflows/doxygen.yaml
@@ -29,6 +29,11 @@ jobs:
                git clone --branch=master --single-branch --depth 1 https://github.com/devkitPro/3ds-examples examples
                cd libctru
                CTRU_VERSION=${{ steps.vars.outputs.tag }} doxygen Doxyfile
+      - name: Publish docs to GH Actions
+        uses: actions/upload-artifact@v2
+        with:
+          path: libctru/docs/*
+          name: docs
 
       - name: Deploy 🚀
         uses: JamesIves/github-pages-deploy-action@3.7.1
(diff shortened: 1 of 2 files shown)


oreo639:
I don't see the point in uploading the docs as CI artifacts (which expires after 90 days), you are probably better off just downloading a zip of the gh-pages branch.

What is docset? The docsetutil repo you are using says it is a re-upload of the docsetutil binary from Xcode, which afaict was marked as deprecated in 8.3 (and removed in 9.3):
https://developer.apple.com/library/archive/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html#//apple_ref/doc/uid/TP40001051

dimaguy:
> I don't see the point in uploading the docs as CI artifacts (which expires after at most 90 days), you are probably better off just downloading a zip of the gh-pages branch.
> 
> What is docset? The docsetutil repo you are using says it is a re-upload of the docsetutil binary from Xcode, which afaict was marked as deprecated in 8.3 (and removed in 9.3):
> https://developer.apple.com/library/archive/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html#//apple_ref/doc/uid/TP40001051

You're kinda correct in both affirmations, I wanted to leave to you however you want to distribute the docset so I made it a minimal artifact thing, but I can easily make it create a release with the docset.  

In regards to the docset, while docsetutil is deprecated the format itself is used by several offline documentation utilities(Zeal, Dash and others), the format bundles the html files with a sqlite database for indexing(improved search) and the utility merely makes it easier to get the job done.

WinterMute:
Using a deprecated tool for macOS in github workflows isn't something we want. If there's other more available tooling for this then maybe but currently no.

PR #523, opened 2023-05-02
Add timeout versions for LightLock, RecursiveLock, and LightSemaphore

piepie62:


libctru/source/synchronization.c
```diff
@@ -82,6 +90,72 @@ void LightLock_Lock(LightLock* lock)
...
+			val = -val; // transition into locked state
+		else
+			--val; // increment the number of waiting threads (which has the sign reversed during locked state)
+	} while (__strex(lock, val));
+
+	// If the lock is held by a different thread, we need to do some timing initialization
+	if (bAlreadyLocked)
+	{
+		struct timespec startTime;
+		clock_gettime(CLOCK_MONOTONIC, &startTime);
```
fincs, line 117: Prefer to use native HOS calls instead of libc from within system support code (applies to the entire file)

fincs:
I'm curious as to which use case prompted this PR.

xzn:
Some questions:

I have a slightly different implementation a while back: [source](https://github.com/xzn/ntr-hr/blob/3a9d9841b21006d298147c7f624a04a418184d45/source/rp/rp_syn.c#L3)

It doesn't yet have decreasing timeout on looped arbitrate attempts, which I tried to remedy in [this version](https://github.com/xzn/libctru/blob/83ed338ac690160dbf2a8c321f269c9a5092e0d7/libctru/source/synchronization.c#L121)

There's a few other differences though:

- I used `ARBITRATION_WAIT_IF_LESS_THAN_TIMEOUT` instead of `ARBITRATION_WAIT_IF_LESS_THAN`, not sure if it matters.
- On timeout or error, for LightLock, the waiting thread count is released by 1 before returning: [source](https://github.com/xzn/libctru/blob/83ed338ac690160dbf2a8c321f269c9a5092e0d7/libctru/source/synchronization.c#L154), I don't see similar code in your PR.
- Similarly for LightSemaphore, the number of threads acquiring the lock is released by 1 before returning: [source](https://github.com/xzn/libctru/blob/83ed338ac690160dbf2a8c321f269c9a5092e0d7/libctru/source/synchronization.c#L531). Again In your PR the status is returned _before_ doing the associated adjustment.

So the question:

Are those adjustment unnecessary (or incorrect) in my version? I do used those code in a homebrew and it works afaict..

@fincs 

As for use case, I just wanted something that matches the behavior of `svcWaitSynchronization` as I have macro to switch between either svc sync or libctru sync for testing purpose.

PR #531, opened 2023-09-11
Do not expose public macros with common names

### LINKED ISSUE #530: Global macros clash with other things
Here, libctru defines some macros that can clash with other things:

https://github.com/devkitPro/libctru/blob/a4634c02905a824d01feae338d1c472ab5dae1eb/libctru/include/3ds/types.h#L51-L52

For example, with the latest version of libfmt, this does not work:

```c++
#include <3ds.h>
#include <fmt/core.h>
```

This is because libfmt uses `PACKED` as a template parameter name in one of its functions:

https://github.com/fmtlib/fmt/blob/d4987546a4269cc21565ea633eed56db09039d1a/include/fmt/core.h#L1557-L1558

Including things in the opposite order works but really 3ds should not define such macros publicly:

```c++
#include <fmt/core.h>
#include <3ds.h>
```

This can be tricky to work around in a medium size codebase, e.g. https://github.com/diasurgical/devilutionX/pull/6585

glebm:
The prevents compatibility issues with other libraries, such as the one seen in #530

Fixes #530


glebm:
Well, my PR to libfmt also got closed https://github.com/fmtlib/fmt/pull/3633 😅
@vitaut

glebm:
@fincs Would you accept a PR that removes or prefixes macros other than `BIT`?

WinterMute:
> Well, my PR to libfmt also got closed [fmtlib/fmt#3633](https://github.com/fmtlib/fmt/pull/3633) 😅 @vitaut

Well, what you can do is just not include libctru headers in the same file as libfmt headers. You might want to point out to libfmt maintainers that "PACKED" is most commonly used for our precise use case & not theirs. See <https://github.com/search?q=%22%23define+PACKED+%22&type=code> for instance.

It seems rather late in the day to be making intrusive & breaking changes like this.

PR #532, opened 2023-09-12
Add CTR_ prefix to ALIGN,PACKED,DEPRECATED macros

### LINKED ISSUE #530: Global macros clash with other things

glebm:
The mitigates compatibility issues with other libraries, such as the one seen in #530.

The `BIT` macro is kept because it is commonly used, so renaming it would be too breaking of a change (see
https://github.com/devkitPro/libctru/pull/531#issuecomment-1714148532)

Fixes #530

### DIFF
--- a/libctru/Doxyfile
+++ b/libctru/Doxyfile
@@ -2023,7 +2023,7 @@ INCLUDE_FILE_PATTERNS  =
 # recursively expanded use the := operator instead of the = operator.
 # This tag requires that the tag ENABLE_PREPROCESSING is set to YES.
 
-PREDEFINED             = LIBCTRU_PACKED
+PREDEFINED             = CTR_PACKED
 
 # If the MACRO_EXPANSION and EXPAND_ONLY_PREDEF tags are set to YES then this
 # tag can be used to specify a list of macro names that should be expanded. The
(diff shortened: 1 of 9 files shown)


WinterMute:
We're leaning towards possibly CTR_ rather than LIBCTRU_

Any opinions?

fincs:
If the prefix introduced here is renamed to `CTR_`, I'd say this PR can be merged.

PR #536, opened 2024-01-07
Fix DmaConfig related flags and member order in structure

xzn:
SRC and DST in DmaConfig related flags and structure seem to be reversed.

Source:

Empirical testing.

(Unless I have a reversed understanding of SRC and DST) [Corrected example](https://github.com/xzn/ntr-hr/blob/04262143c75c4195c095e53f8d749b248ada75b9/source/ns/ns.c#L1550)

Note: this change is NOT backward compatible, assuming there are existing code relying on the reversed behavior.

### DIFF
--- a/libctru/include/3ds/svc.h
+++ b/libctru/include/3ds/svc.h
@@ -148,12 +148,12 @@ typedef enum {
 
 /// Configuration flags for \ref DmaConfig.
 enum {
-	DMACFG_SRC_IS_DEVICE        = BIT(0), ///< DMA source is a device/peripheral. Address will not auto-increment.
-	DMACFG_DST_IS_DEVICE        = BIT(1), ///< DMA destination is a device/peripheral. Address will not auto-increment.
+	DMACFG_DST_IS_DEVICE        = BIT(0), ///< DMA destination is a device/peripheral. Address will not auto-increment.
+	DMACFG_SRC_IS_DEVICE        = BIT(1), ///< DMA source is a device/peripheral. Address will not auto-increment.
 	DMACFG_WAIT_AVAILABLE       = BIT(2), ///< Make \ref svcStartInterProcessDma wait for the channel to be unlocked.
 	DMACFG_KEEP_LOCKED          = BIT(3), ///< Keep the channel locked after the transfer. Required for \ref svcRestartDma.
-	DMACFG_USE_SRC_CONFIG       = BIT(6), ///< Use the provided source device configuration even if the DMA source is not a device.
-	DMACFG_USE_DST_CONFIG       = BIT(7), ///< Use the provided destination device configuration even if the DMA destination is not a device.
+	DMACFG_USE_DST_CONFIG       = BIT(6), ///< Use the provided destination device configuration even if the DMA destination is not a device.
+	DMACFG_USE_SRC_CONFIG       = BIT(7), ///< Use the provided source device configuration even if the DMA source is not a device.
 };
 
 /// Configuration flags for \ref svcRestartDma.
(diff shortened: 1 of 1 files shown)


TuxSH:
I have checked my k11 idb, the current code is correct

xzn:
@TuxSH Don't mean to spam this but could you provide more information?

In my usage of `svcStartInterProcessDma`, I have to put the pitch of destination memory in the first `DmaDeviceConfig` while pitch of the src memory in the second `DmaDeviceConfig` in `DmaConfig` before it works as expected, which is why I think its reversed.

TuxSH:
I have checked in my kernel disassembly that the correct flag and substructure was used for the src/dst addresses, respectively.

The whole thing is a bit confusing but IIRC the burst/transfer thing is there because the DMA controller has two loop counters. The "burst" parameters are meant to be used for the inner loop while the "transfer" parameters meant to be used for the outer one.

xzn:
Hmm I may look into a k11 dump with Ghidra when I have time (just to figure out what's going on).

Nevertheless, I'm actually using `transferStride` as the pitch for src and dst address transfer, and as I've mentioned the src pitch goes into the second `DmaDeviceConfig` struct, not the first, and vice versa for dst. I've tried tuning a couple of other parameters in `DmaConfig` and they all point to the same thing. Are you sure we are looking at/talking about the same thing?

If you have time I can try to get a minimal homebrew sample to demonstrate the issue. Would be easier for me than to figure out where to get the k11 dump haha.

TuxSH:
The structure and flag enum are both correctly defined.

Can you try setting burstSize and burstStride to 16 and 16, for both dst and src?

xzn:
Just tried it and it didn't make a difference. I still need to set the pitch of src buffer into the second config struct in order to make it work:

https://github.com/xzn/ntr-hr/blob/7db5dfdd1b7c045e227b603cc4e1d6b7661bdfc6/source/nwm/rp_nwm.c#L1034

(Note: I'm using adjusted version of libctru in the above code, so src config is the second config struct there.)

My updated code for testing:
```c
...

	u32 bpp; /* bytes per pixel */
	u32 burstSize = 16; /* highest power-of-2 factor of 240 */

	if (format == 0){
		bpp = 4;
	}
	else if (format == 1){
		bpp = 3;
	}
	else{
		bpp = 2;
	}
	u32 transferSize = 240 * bpp;
	u32 pitch = isTop ? tl_pitch : bl_pitch;

	u32 bufSize = transferSize * (isTop ? 400 : 320);

	DmaConfig dmaConfig = {
		.channelId = -1,
		.flags = DMACFG_WAIT_AVAILABLE | DMACFG_USE_DST_CONFIG | DMACFG_USE_SRC_CONFIG,
		.dstCfg = {
			.deviceId = -1,
			.allowedAlignments = 15,
			.burstSize = burstSize,
			.burstStride = burstSize,
			.transferSize = transferSize,
			.transferStride = transferSize,
		},
		.srcCfg = {
			.deviceId = -1,
			.allowedAlignments = 15,
			.burstSize = burstSize,
			.burstStride = burstSize,
			.transferSize = transferSize,
			.transferStride = pitch,
		},
	};
...
```

Pitch is set from here:
https://github.com/xzn/ntr-hr/blob/7db5dfdd1b7c045e227b603cc4e1d6b7661bdfc6/source/nwm/rp_nwm.c#L1098

When using say DaedalusX64, top screen has 800 columns of pixels. and I used the pitch parameter to skip every other column when doing DMA.

WinterMute:
> Just tried it and it didn't make a difference. I still need to set the pitch of src buffer into the second config struct in order to make it work:
> 
> https://github.com/xzn/ntr-hr/blob/7db5dfdd1b7c045e227b603cc4e1d6b7661bdfc6/source/nwm/rp_nwm.c#L1034
> 
> (Note: I'm using adjusted version of libctru in the above code, so src config is the second config struct there.)

When you have a testcase built with stock tools using stock libcrtu then do feel free to open a new issue. We can't provide support for this kind of project, sorry.

PR #541, opened 2024-03-19
Implement linearRealloc

BallM4788:


BallM4788:
@WinterMute Would you mind checking this out?

PR #542, opened 2024-05-04
Fix socket poll events values

DaniElectra:
The socket sysmodule for the 3DS uses the same codebase as the Wii sockets. We already know the the [Wii values](https://github.com/devkitPro/libogc/blob/0ea72783a3792ed7d9fa2571f8f2f66676467099/gc/network.h#L198-L207) for poll events, and the 3DS is no exception to these. These values were already documented on [3dbrew](https://www.3dbrew.org/wiki/SOCU:poll), but libctru was still using the wrong values.

These wrong values had gone unnoticed because the read and write events were pointing to *somewhat* correct values, and wrong values for errors weren't triggered because applications don't set the error events on input, assuming that the system would trigger the error values even if it isn't given as input. This is correct, but since the error values were pointing to incorrect places, this wasn't noticed.

The values have been verified when reverse engineering the socket sysmodule, and has been tested with various applications:
- mbedtls
- curl
- Luma3DS (gdb stub)
- ftpd


mtheall:
This doesn't work for POLLPRI
When data comes over OOB then 0x1 and 0x2 are triggered but not 0x4

This means it breaks ftpd's transfer cancellation mechanism

The wiki says:
`POLLPRI: Urgent data can be read.`
but I have verified that 0x2 is triggered (not 0x4) when the TCP message has the URGENT flag set

mtheall:
I think it would probably be acceptable if
```cpp
#define POLLPRI (POLLRDBAND | 0x04)
```

DaniElectra:
Upon further inspection I can't find any path in the code that can trigger 0x04, 0x40, or 0x80. POLLNVAL is also 0x20 indeed, that is my bad.

The write flags are probably right, since when flag 0x10 is set, flag 0x8 is always set too. And on UDP the only revents returned are either 0x1 (POLLIN) or 0x9 (POLLIN | POLLWRBAND)

Surprisingly, the module adds `0xE0` on the input event flags even though there doesn't seem to be a path to trigger 0x40 or 0x80

TuxSH:
Sorry if this hasn't been responded to yet -- I'll try to have a look again this week

PR #550, opened 2024-09-09
Return the full length of `sockaddr_in` in socket functions with in/out addrlen

Meziu:
Relevant links: https://github.com/rust3ds/ctru-rs/issues/174, [discussion on courses of action](https://github.com/Meziu/rust-horizon/commit/96b0f4c2d3f61863b1499037d3a738f73cb4511a#r145611578), [original change in libc crate](https://github.com/rust-lang/libc/pull/2725), [problematic test](https://github.com/rust-lang/rust/blob/38e3a5771cefc9362976a605549f8b04d5707311/library/std/src/sys_common/net.rs#L113).
CC: @ian-h-chamberlain @adryzz @AzureMarker

The kernel implementation of `sockaddr` data, used by all `soc` services, is composed of 8 bytes of data. While this is true, the `sockaddr_in` definition exposed by libctru is more closely related to the corresponding Linux (and many other systems') definition, having an additional 8 byte padding field (`sin_zero`). This is a generally favorable design decision for compatibility with most *nix software. However, unlike most Posix-complying implementations, the current implementation in `libctru` only handles the transfer of the 8 bytes actually manipulated by the OS.

This particularity imposes 2 "differentiating" factors that may hinder compatiblity with code written for standard *nix platforms:
1. The padding field is never (internally) overwritten to be actually zeroed. In most situations that means that programs passing in a pointer to `sockaddr_in` memory are left with uninitialized data (not a big deal since that memory isn't responsibility of `libctru`). This also shouldn't be a problem since the implementation of functions with an in/out `*addrlen` argument correctly specify that only the first 8 bytes were modified (also, `sin_zero` isn't supposed to ever be read).
2. The returned length, as specified above being always 8 bytes, is less than the size of the `sockaddr_in` structure (16 bytes), meaning that programs expecting the size given to be the same as the output will be unpleasantly surprised when they find out the `sockaddr_in` struct is "unexpectedly" oversized. For this reason (and for the fact that `sin_zero` is a fundamentally important field for most implementations), platforms such as Linux actually return the full 16 byte length, taking into consideration those 8 trailing bytes. This isn't done by `libctru`, which only returns a socket address length of 8.

I am fully aware that this is basically a non-existent problem, and as far as I can tell, the Posix standard isn't even broken by having an oversized `sockaddr_in` definition, since data length issues are to be resolved using the `*addrlen` parameter, and not by "expecting" sizes and lengths to be that of data structures. However, and the case that prompted this change is proof of this, this situation is not really expected to be handled by most programs, which instead believe the platform's socket implementation has to account for the full memory mapping (which is something they usually do).

The possible fixes for this problem are:
- Do nothing about it. (not cool)
- Remove the `sin_zero` field from the `libctru` definition and more thinly wrap the kernel implementation. (might break backward and cross compatibility)
- Return the full length of `sockaddr_in` for successful calls to socket functions. Since this is the solution which requires the most changes, I've taken the liberty to include such changes in this PR.

This PR doesn't have to be merged or closed as-is. Please take the time to discuss the issue further if this is not what you intend on using.

### DIFF
--- a/libctru/source/services/soc/soc_common.h
+++ b/libctru/source/services/soc/soc_common.h
@@ -10,6 +10,7 @@
 #include <3ds/services/soc.h>
 
 #define SYNC_ERROR ENODEV
+#define ADDR_STORAGE_LEN sizeof(struct sockaddr_storage)
 
 extern Handle	SOCU_handle;
 extern Handle	socMemhandle;
--- a/libctru/source/services/soc/soc_accept.c
+++ b/libctru/source/services/soc/soc_accept.c
@@ -61,9 +61,14 @@ int accept(int sockfd, struct sockaddr *addr, socklen_t *addrlen)
 
 	if(ret >= 0 && addr != NULL) {
 		addr->sa_family = tmpaddr[1];
-		if(*addrlen > tmpaddr[0])
-			*addrlen = tmpaddr[0];
-		memcpy(addr->sa_data, &tmpaddr[2], *addrlen - 2);
+
+		socklen_t user_addrlen = tmpaddr[0];
+		if(addr->sa_family == AF_INET)
+		    user_addrlen += 8; // Accounting for the 8 bytes of sin_zero padding, which must be written for compatibility.
+
+		if(*addrlen > user_addrlen)
+			*addrlen = user_addrlen;
+		memcpy(addr->sa_data, &tmpaddr[2], *addrlen - sizeof(addr->sa_family));
 	}
 
 	if(ret < 0) {
--- a/libctru/source/services/soc/soc_accept.c
+++ b/libctru/source/services/soc/soc_accept.c
@@ -6,10 +6,10 @@
 int accept(int sockfd, struct sockaddr *addr, socklen_t *addrlen)
 {
 	int ret = 0;
-	int tmp_addrlen = 0x1c;
+	int tmp_addrlen = ADDR_STORAGE_LEN;
 	int fd, dev;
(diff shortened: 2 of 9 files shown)


mtheall:
Can you split the commit into one that updates all the 0x1c to ADDR_STORAGE_LEN and one that does the other stuff?

libctru/source/services/soc/soc_common.h
```diff
@@ -10,6 +10,7 @@
 #include <3ds/services/soc.h>
 
 #define SYNC_ERROR ENODEV
+#define ADDR_STORAGE_LEN sizeof(struct sockaddr_storage)
```
mtheall, line 13: Maybe we should have a `_Static_assert(ADDR_STORAGE_LEN == 0x1c);`

mtheall (changes requested):

Meziu:
> Can you split the commit into one that updates all the 0x1c to ADDR_STORAGE_LEN and one that does the other stuff?

They are already separated 👍

libctru/source/services/soc/soc_common.h
Meziu: My idea was to not hardcode the size to any particular value (this PR is of the intent to revise the size of `sockaddr` after all, and I remember that the 3DS doesn’t have IPv6 support). It could be a helpful test though, I’ll see to add it if I have to make any changes.

sardap:
It would be great if this could get some movement!

PR #554, opened 2024-11-20
[GPU] Initialize interrupt queue header to 0

### LINKED ISSUE #521: If homebrew is launched directly after NSMB 2, GPU commands aren't being executed anymore
To reproduce this bug just launch New Super Mario Bros 2, and afterwards any other homebrew game. Now, after initializing the GPU, no gpu commands (like `GX_MemoryFill`) seem to go through. The functions also never return any errors. And as soon as you start doing more stuff, like drawing a frame, the gpu crashes completely. Despite trying for quite a while I wasn't able to figure out how to fix the crash, I simply don't know enough about how the GPU works.
While looking into the source code I noticed another oddity by the way. Calling `GX_MemoryFill` and only using the 2nd buffer doesn't work, as the event for a single memory fill being done on channel two is surpressed in `source/gpu/gxqueue.c`, line 29. Removing the check for that specific event doesn't seem to affect anything, though I haven't thouroughly tested that and I might be missing something. This doesn't matter much, only if the first buffer is busy for some reason so you want to use the second one.

kynex7510:
Fixes #521.

### DIFF
--- a/libctru/source/services/gspgpu.c
+++ b/libctru/source/services/gspgpu.c
@@ -184,6 +184,12 @@ Result gspInit(void)
 	gspSharedMem = mappableAlloc(0x1000);
 	svcMapMemoryBlock(gspSharedMemHandle, (u32)gspSharedMem, MEMPERM_READWRITE, MEMPERM_DONTCARE);
 
+	// Initialize interrupt queue header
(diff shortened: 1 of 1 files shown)


fincs:
Thank you very much for taking the time to debug, investigate and fix this issue, much appreciated :)

catfromplan9:
Can a new release including this be made?

PR #557, opened 2025-02-09
apt: Add aptLaunchSystemApplet

DaniElectra:
Needed for launching system applets like the Miiverse posting applet "solv3". The screen transfer is expected to be done after launching the applet.


fincs:
I believe only the aptScreenTransfer needs to be done after APT_StartSystemApplet; the other two (GSPGPU_SaveVramSysArea/GSPGPU_ReleaseRight) should still be done before; as in aptLaunchLibraryApplet.

DaniElectra:
You are right, that should be fixed now

PR #558, opened 2025-03-06
Implement ACT, finalize FRD implementation, fix Mii struct

ZeroSkill1:
- Implemented the entirety of the friends services commands, fixing incorrect ones and adding all the missing ones.
- Fix up the existing base Mii struct.

### DIFF
--- a/libctru/include/3ds/mii.h
+++ b/libctru/include/3ds/mii.h
@@ -6,12 +6,16 @@
  */
 #pragma once
 
+#define MII_NAME_LEN           10+1  ///< 10-character NULL-terminated UTF-16 mii name
+
+typedef u16 MiiScreenName[MII_NAME_LEN];
+
 #include <3ds/types.h>
 
-/// Shared Mii struct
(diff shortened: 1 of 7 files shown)


TuxSH:
Looks good, though, could you give more context on why you added PACKED?

fincs:
Some of the changes add PACKED to structures that only contain 8-bit members. Why is that?

ZeroSkill1:
That must've been an oversight from when I was trying to fix up the incorrect alignments. I've removed CTR_PACKED from the inner structs that don't need it.

ZeroSkill1:
While I'm at it, I'm also going to add the implementation for ACT services.

fincs:
Is this PR done? In any case, would you mind tidying up history a bit once you're done? Ideally there should be one commit per service.

ZeroSkill1:
Yes, I've added everything I wanted to change.

I've reorganized the the commits so each change (Mii struct, FRD, ACT) is its own separate commit.

PR #559, opened 2025-03-07
acu: Added GetProxyHost

R-YaTian:
Also: change output buffer to `u16*` for `ACU_GetProxyPort`.


fincs:
Ideally, the size of output buffers should be included in the docstring. Would you mind adding them?

R-YaTian:
> Ideally, the size of output buffers should be included in the docstring. Would you mind adding them?

Done!

PR #562, opened 2025-03-16
fix: Sleep and wake up dsp when calling a system applet

Sorunome:
In https://github.com/devkitPro/libctru/commit/a88ab665a83fd15483b4da9760631965ebeef092 a new method `aptLaunchSystemApplet` was added. This, however, does not pause / wake up the DSP, if it is in use. This is required, else the applet may do a fantastic crash (e.g. browser applet).

This PR fixes this by calling the respective methods.

This has been tested with the browser applet both while using and while not using dsp.

### DIFF
--- a/libctru/source/services/apt.c
+++ b/libctru/source/services/apt.c
@@ -921,6 +921,7 @@ void aptLaunchSystemApplet(NS_APPID appId, void* buf, size_t bufsize, Handle han
 	aptCallHook(APTHOOK_ONSUSPEND);
 
 	GSPGPU_SaveVramSysArea();
+	aptDspSleep();
 	GSPGPU_ReleaseRight();
 
 	aptSetSleepAllowed(false);


libctru/source/services/apt.c
```diff
@@ -911,6 +911,7 @@ void aptLaunchLibraryApplet(NS_APPID appId, void* buf, size_t bufsize, Handle ha
 
 void aptLaunchSystemApplet(NS_APPID appId, void* buf, size_t bufsize, Handle handle)
 {
+	aptDspSleep();
```
fincs, line 914: I believe the call to aptDspSleep should be placed between GSPGPU_SaveVramSysArea and GSPGPU_ReleaseRight, just like in aptJumpToHomeMenu.

libctru/source/services/apt.c
```diff
@@ -932,6 +933,7 @@ void aptLaunchSystemApplet(NS_APPID appId, void* buf, size_t bufsize, Handle han
 	aptWaitForWakeUp(TR_SYSAPPLET);
 	memcpy(buf, aptParameters, bufsize);
 	aptSetSleepAllowed(sleep);
+	aptDspWakeup();
```
fincs, line 936: Likewise, I believe this call should be removed as in aptJumpToHomeMenu, because it is part of aptWaitForWakeUp.

libctru/source/services/apt.c
Sorunome: you are right, fixed (tested and pushed) this now

libctru/source/services/apt.c
Sorunome: see above

Sorunome:
Thank you for the merge! Soru was wondering, though, is there any reason you made this a different commit? Soru likes to sign her code to have proof of having written it, which is now gone with the way you merged. Not that it is the end of the world, but it would be, like, nice to preserve that

fincs:
I used GitHub's rebase & merge function. Merge commits are disallowed.

Sorunome:
huh, soru thought it was rebased to master's head. oh well, thank you!

PR #565, opened 2025-03-19
Return the full length of `sockaddr_in` in `getaddrinfo`

DCNick3:
In #550, `getaddrinfo` was left out of the changes to returned lengths of the socket addresses. This caused the same asserts in Rust when trying to resolve domain names.

This PR makes gives the same treatment to `getaddrinfo`, making the reported length include `sin_zero`.

### DIFF
--- a/libctru/source/services/soc/soc_getaddrinfo.c
+++ b/libctru/source/services/soc/soc_getaddrinfo.c
@@ -53,6 +53,14 @@ static struct addrinfo * buffer2addrinfo(addrinfo_3ds_t * entry)
 
 		memcpy(ai->ai_canonname, entry->ai_canonname, ai_canonname_len);
 		memcpy(ai->ai_addr, &entry->ai_addr, ai->ai_addrlen);
+
+		// the sizes of hos-native sock_addr_in and the one we expose from libc are not the same
+		// because the libc one has a `sin_zero` member for better compatibility with linux
+		// it is located at the tail of the structure and will be zero-initilized due to use of `calloc` above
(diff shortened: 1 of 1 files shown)


libctru/source/services/soc/soc_getaddrinfo.c
```diff
@@ -52,7 +52,14 @@ static struct addrinfo * buffer2addrinfo(addrinfo_3ds_t * entry)
 		ai->ai_addrlen  = entry->ai_addrlen;
 
 		memcpy(ai->ai_canonname, entry->ai_canonname, ai_canonname_len);
+
+		// make sure we account for sin_zero, which the hos-native sockaddr_in doesn't have
+		memset(ai->ai_addr, 0, sizeof(struct sockaddr_storage));
 		memcpy(ai->ai_addr, &entry->ai_addr, ai->ai_addrlen);
+
+		if(ai->ai_family == AF_INET)
+			ai->ai_addrlen += 8;
```
fincs, line 61: Is there any reason why this can't just be `ai->ai_addrlen = sizeof(struct sockaddr_in)`?

DCNick3: I don't see any reason it can't be that.

I just did it the same way #550 did it in several places. Should I then change those to `= sizeof(struct sockaddr_in)` for consistency, or only update it in here?

fincs: We should probably change the other functions as well in order to make the code clearer & not use hardcoded constants. But for the sake of this PR, I think it's OK to only limit the change to this function.

PR #567, opened 2025-03-24
Add root makefile

impeeza:
Add a simple makefile on the root of the repository which call the correct makefile on the libctru subfolder.


WinterMute:
This isn't a simple Makefile and the repository already has one in the libcrtu folder.

impeeza:
That was the idea, almost all repositories have a makefile on the root so you do not need to go to a subdirectory to build it, this makefile is more like a shortcut.  Regards

PR #568, opened 2025-03-31
Add (incomplete) ir:USER API based on ctru-rs implementation for Circle Pad Pro support

syelan34:
Adds basic API similar to the ir:U and ir:RST APIs, with [all commands documented on 3DBrew](https://www.3dbrew.org/wiki/IR_Services#ir:USER) implemented. Implementation of some functions was based on [that of ctru-rs](https://github.com/rust3ds/ctru-rs/blob/master/ctru-rs/src/services/ir_user.rs).

I don't have a circle pad pro to test with so I'm unable to check if everything works correctly, but it should match any documentation that exists.

### DIFF
--- a/libctru/source/services/iruser.c
+++ b/libctru/source/services/iruser.c
@@ -486,10 +486,10 @@ static bool iruserParsePacket(size_t index, IRUSER_Packet* packet) {
     
     u32 payload_offset = 0;
     if (large) {
-        packet->payload_length = ((*IRUSER_PACKET_DATA(inf.offset + 2) & 0x4F) << 8) + *IRUSER_PACKET_DATA(inf.offset + 3);
+        packet->payload_length = ((*IRUSER_PACKET_DATA(inf.offset + 2) & 0x3F) << 8) | *IRUSER_PACKET_DATA(inf.offset + 3);
         payload_offset = 4;
     } else {
-        packet->payload_length = *IRUSER_PACKET_DATA(inf.offset + 2) & 0x4F; // bottom 6 bits
+        packet->payload_length = *IRUSER_PACKET_DATA(inf.offset + 2) & 0x3F; // bottom 6 bits
         payload_offset = 3;
     }
     


syelan34:
I'm not super satisfied with my implementation of the iruserGetPackets() function, since it allocates two separate buffers, so please suggest any alternatives which might match how other libctru functions might act.
The variable length encoding of packets, as well as the unknown number of packets makes it difficult to do any other way which is why I haven't so far.

libctru/source/services/iruser.c
```diff
@@ -0,0 +1,569 @@
...
+    0x76, 0x71, 0x78, 0x7F, 0x6A, 0x6D, 0x64, 0x63,
+    0x3E, 0x39, 0x30, 0x37, 0x22, 0x25, 0x2C, 0x2B,
+    0x06, 0x01, 0x08, 0x0F, 0x1A, 0x1D, 0x14, 0x13,
+    0xAE, 0xA9, 0xA0, 0xA7, 0xB2, 0xB5, 0xBC, 0xBB,
+    0x96, 0x91, 0x98, 0x9F, 0x8A, 0x8D, 0x84, 0x83,
+    0xDE, 0xD9, 0xD0, 0xD7, 0xC2, 0xC5, 0xCC, 0xCB,
+    0xE6, 0xE1, 0xE8, 0xEF, 0xFA, 0xFD, 0xF4, 0xF3
+};
+
+u8 crc8ccitt(const void * data, size_t size) {
```
TuxSH, line 70: static?

libctru/source/services/iruser.c
```diff
@@ -0,0 +1,569 @@
...
+
+/// Read and parse the current packets received from the IR device.
+IRUSER_Packet* iruserGetPackets(u32* n) {
+    void* shared_mem = iruserSharedMem;
+
+    // Find where the packets are, and how many
+    IRUSER_BufferInfo buffer_info;
+    memcpy(&buffer_info, (u8*)shared_mem + 0x10, sizeof(buffer_info));
+    
+    if (n) *n = buffer_info.valid_packet_count;
```
TuxSH, line 529: You dereference n unconditionally below, please check

libctru/source/services/iruser.c
```diff
@@ -0,0 +1,569 @@
...
+    if (R_FAILED(ret)) goto cleanup1;
+
+    ret = IRUSER_InitializeIrNopShared(buffer_size, packet_count, buffer_size, packet_count, IR_BITRATE);
+    if (R_FAILED(ret)) goto cleanup2;
+
+    iruserRecvBufferSize = buffer_size;
+    iruserRecvPacketCount = packet_count;
+    iruserRecvPacketInfoBuffer = (IRUSER_PacketInfo*)((u8*)iruserSharedMem + SHARED_MEM_RECV_BUFFER_OFFSET);
+    iruserRecvPacketDataBuffer = (u8*)iruserRecvPacketInfoBuffer + iruserRecvPacketCount * PACKET_INFO_SIZE;
+    iruserRecvPacketDataBufferSize = (u32)((u8*)iruserSharedMem - iruserRecvPacketDataBuffer);
```
TuxSH, line 104: This doesn't look right,

Shouldn't the offset be the other way round or equal to ((u8*)iruserRecvPacketDataBuffer - (u8*)iruserSharedMem)? Please double read and check

libctru/source/services/iruser.c
```diff
@@ -0,0 +1,569 @@
...
+        packet->payload_length = *IRUSER_PACKET_DATA(2);
+        packet->payload = malloc(packet->payload_length);
+        payload_offset = 3;
+    }
+    for (size_t i = 0; i < packet->payload_length; i++) {
+        packet->payload[i] = *IRUSER_PACKET_DATA(i + payload_offset);
+    }
+    packet->checksum = *IRUSER_PACKET_DATA(packet->payload_length + payload_offset);
+    // check the checksum
+    u8 checksum = crc8ccitt(IRUSER_PACKET_DATA(0), packet->payload_length + payload_offset);
```
TuxSH, line 508: bogus, should be u8 checksum = crc8ccitt(packet->payload, packet->payload_length);

libctru/source/services/iruser.c
```diff
@@ -0,0 +1,569 @@
...
+    0xAE, 0xA9, 0xA0, 0xA7, 0xB2, 0xB5, 0xBC, 0xBB,
+    0x96, 0x91, 0x98, 0x9F, 0x8A, 0x8D, 0x84, 0x83,
+    0xDE, 0xD9, 0xD0, 0xD7, 0xC2, 0xC5, 0xCC, 0xCB,
+    0xE6, 0xE1, 0xE8, 0xEF, 0xFA, 0xFD, 0xF4, 0xF3
+};
+
+u8 crc8ccitt(const void * data, size_t size) {
+	u8 val = 0;
+
+	u8 * pos = (u8 *) data;
```
TuxSH, line 73: const u8

libctru/source/services/iruser.c
```diff
@@ -0,0 +1,569 @@
...
+    // Result ret = 0;
+    u32 n = 0;
+    IRUSER_Packet* packet = iruserGetPackets(&n);
+    if (n == 0) return -1;
+    if (packet[n-1].payload_length != 6) return RS_INVALIDSTATE;
+    if (packet[n-1].payload[0] != CIRCLE_PAD_PRO_INPUT_RESPONSE_PACKET_ID) return RS_INVALIDSTATE;
+    pos->dx = (s16)(packet[n-1].payload[1] | ((packet[n-1].payload[2] & 0x0F) << 8));
+    pos->dy = (s16)(((packet[n-1].payload[2] & 0xF0) >> 4) | ((packet[n-1].payload[3]) << 4));
+
+    free(packet[n-1].payload);
```
TuxSH, line 477: Many memory leaks in this function (leaks all n > 1, and the early return also leak)

TuxSH (changes requested):

TuxSH:
Forgot to add review comment:

I think CPP support is good to have (though ideally this would need to be better integrated with hid/ir key sampling code).

There are a few critical bugs with your impl (notably with regards to lifetime and memory ownership).

I wonder if the API could be redesigned such that user provides user buffer instead of it being allocated. I haven't checked if this is possible, but dynamic allocations should be avoided there are cheap alternatives.

PR #569, opened 2025-04-03
Implemented mvdstdCalculateBufferSize()

Core-2-Extreme:
I've [analyzed](https://www.3dbrew.org/wiki/MVDSTD:CalculateWorkBufSize) and implemented `mvdstdCalculateBufferSize()` to calculate working buffer size for H.264 video decoding on MVD based on video (resolution/H.264 level etc...).

Currently we don't have any way to calculate working buffer size, so only way to use MVD is just use `MVD_DEFAULT_WORKBUF_SIZE` (or ~~maybe just put random value to check if it works for your video~~), it's fine if you want to decode (most of) 480p videos.

However that doesn't mean MVD can't decode more than that (actually it **can** decode 1080p if enough buffer is provided) so we should also have a way to calculate it, this can also be used to calculate smaller size for lower resolution/level videos.

### DIFF
--- a/libctru/include/3ds/services/mvd.h
+++ b/libctru/include/3ds/services/mvd.h
@@ -167,7 +167,7 @@ void mvdstdExit(void);
 
 /**
  * @brief Calculate working buffer size for H.264 decoding.
- * @param config Calculation config, see here for detailed explanations : https://www.3dbrew.org/wiki/MVDSTD:CalculateWorkBufSize.
+ * @param config Calculation config, config->level.level must NOT exceed MVD_H264_LEVEL_5_2. See here for more explanations : https://www.3dbrew.org/wiki/MVDSTD:CalculateWorkBufSize.
  * @param size_out Calculated buffer size in bytes.
  */
 Result mvdstdCalculateBufferSize(const MVDSTD_CalculateWorkBufSizeConfig* config, u32* size_out);
--- a/libctru/include/3ds/services/mvd.h
+++ b/libctru/include/3ds/services/mvd.h
@@ -20,28 +20,28 @@
 /// Default input size for mvdstdInit(). This is what the New3DS Internet Browser uses, from the MVDSTD:CalculateWorkBufSize output.
 #define MVD_DEFAULT_WORKBUF_SIZE 0x9006C8
 
-#define MVD_CALC_WITH_LEVEL_FLAG_NONE				(u8)(0x00)	//Nothing.
-#define MVD_CALC_WITH_LEVEL_FLAG_ENABLE_CALC		(u8)(0x01)	//Enable calculation with level.
-#define MVD_CALC_WITH_LEVEL_FLAG_ENABLE_EXTRA_OP	(u8)(0x02)	//Enable extra op after base calculation (see : https://www.3dbrew.org/wiki/MVDSTD:CalculateWorkBufSize).
-#define MVD_CALC_WITH_LEVEL_FLAG_UNK				(u8)(0x04)	//Unknown.
+#define MVD_CALC_WITH_LEVEL_FLAG_NONE				0x00	//Nothing.
+#define MVD_CALC_WITH_LEVEL_FLAG_ENABLE_CALC		0x01	//Enable calculation with level.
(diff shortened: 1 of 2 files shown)


libctru/include/3ds/services/mvd.h
```diff
@@ -109,6 +165,13 @@ Result mvdstdInit(MVDSTD_Mode mode, MVDSTD_InputFormat input_type, MVDSTD_Output
 /// Shuts down MVDSTD.
 void mvdstdExit(void);
 
+/**
+ * @brief Calculate working buffer size for H.264 decoding.
+ * @param config Calculation config, see here for detailed explanations : https://www.3dbrew.org/wiki/MVDSTD:CalculateWorkBufSize.
+ * @param size_out Calculated buffer size in bytes.
+ */
+Result mvdstdCalculateBufferSize(MVDSTD_CalculateWorkBufSizeConfig* config, u32* size_out);
```
fincs, line 173:
```suggestion
Result mvdstdCalculateBufferSize(const MVDSTD_CalculateWorkBufSizeConfig* config, u32* size_out);
```

libctru/source/services/mvd.c
```diff
@@ -50,6 +52,20 @@ static Result MVDSTD_Shutdown(void)
 	return cmdbuf[1];
 }
 
+static Result MVDSTD_CalculateWorkBufSize(MVDSTD_CalculateWorkBufSizeConfig* config, u32* size_out)
```
fincs, line 55:
```suggestion
static Result MVDSTD_CalculateWorkBufSize(const MVDSTD_CalculateWorkBufSizeConfig* config, u32* size_out)
```

libctru/source/services/mvd.c
```diff
@@ -304,10 +322,45 @@ void mvdstdExit(void)
...
+		if(R_FAILED(ret=srvGetServiceHandle(&mvdstdHandle, "mvd:STD")))
+			return ret;
+
+		must_close_handle = true;
+	}
+
+	//Attempting to use more than 0x10 for H.264 level will likely cause out-of-bound access. 
+	//https://www.3dbrew.org/wiki/MVDSTD:CalculateWorkBufSize
+	if(config->level.level > 0x10)
+		config->level.level = 0x10;
```
fincs, line 350: This should be changed so that it returns an error instead of overwriting an input struct (& document that in the header)

libctru/source/services/mvd.c
```diff
@@ -304,10 +322,45 @@ void mvdstdExit(void)
 	MVDSTD_Shutdown();
 
 	svcCloseHandle(mvdstdHandle);
+	mvdstdHandle = 0;
 
 	linearFree(mvdstd_workbuf);
 }
 
+Result mvdstdCalculateBufferSize(MVDSTD_CalculateWorkBufSizeConfig* config, u32* size_out)
```
fincs, line 330:
```suggestion
Result mvdstdCalculateBufferSize(const MVDSTD_CalculateWorkBufSizeConfig* config, u32* size_out)
```

libctru/include/3ds/services/mvd.h
```diff
@@ -20,6 +20,29 @@
...
+#define MVD_H264_LEVEL_2_2		(u8)(0x07)	//H.264 level 2.2.
+#define MVD_H264_LEVEL_3_0		(u8)(0x08)	//H.264 level 3.0.
+#define MVD_H264_LEVEL_3_1		(u8)(0x09)	//H.264 level 3.1.
+#define MVD_H264_LEVEL_3_2		(u8)(0x0A)	//H.264 level 3.2.
+#define MVD_H264_LEVEL_4_0		(u8)(0x0B)	//H.264 level 4.0.
+#define MVD_H264_LEVEL_4_1		(u8)(0x0C)	//H.264 level 4.1.
+#define MVD_H264_LEVEL_4_2		(u8)(0x0D)	//H.264 level 4.2.
+#define MVD_H264_LEVEL_5_0		(u8)(0x0E)	//H.264 level 5.0.
+#define MVD_H264_LEVEL_5_1		(u8)(0x0F)	//H.264 level 5.1.
+#define MVD_H264_LEVEL_5_2		(u8)(0x10)	//H.264 level 5.2.
```
fincs, line 44: No real need to do `(u8)(xxx)`, just do `xxx`

fincs (changes requested):

PR #577, opened 2025-06-09
Complete AM services implementation

ZeroSkill1:
- Added all missing AM service commands.
- Added missing AM struct / enum definitions.
- Improve existing AM struct definitions.
- Improve overall documentation for commands and structs/enum values.


TuxSH (changes requested):
Other than minor concerns, LGTM

PR #591, opened 2025-09-14
feat: Try to get the dsp firmware from the home menu title, too

Sorunome:
This adds a new way to fetch the dsp firmware, which will work with luma installs.


fincs:
Luma3DS's Rosalina menu already has an option for dumping the DSP firmware binary. libctru isn't really an appropriate place to add DSP firmware dumping code, sorry.

PR #594, opened 2025-11-22
Tiny ps improvement

kynex7510:
Matches the structure described in [3ds_ps](https://github.com/luigoalma/3ds_ps).

### DIFF
--- a/libctru/include/3ds/services/ps.h
+++ b/libctru/include/3ds/services/ps.h
@@ -35,7 +35,8 @@ typedef struct {
 	u8 modulo[0x100];
 	u8 exponent[0x100];
 	u32 rsa_bitsize;//The signature byte size is rsa_bitsize>>3.
-	u32 unk;//Normally zero?
+	u8 is_full_exponent; // 1 if big endian full exponent, 0 if small 4 byte little endian exponent
+	u8 padding[3];
 } psRSAContext;
 
 /// Initializes PS.


libctru/include/3ds/services/ps.h
```diff
@@ -35,7 +35,8 @@ typedef struct {
 	u8 modulo[0x100];
 	u8 exponent[0x100];
 	u32 rsa_bitsize;//The signature byte size is rsa_bitsize>>3.
-	u32 unk;//Normally zero?
+	u8 is_full_exponent; // 1 if big endian full exponent, 0 if small 4 byte little endian exponent
```
TuxSH, line 38: Isn't it likely `is_private_key`? The "small" exponent, is it 0x10001 perchance?

TuxSH (changes requested):
Could you also please rephrase your commit's title? Thanks

libctru/include/3ds/services/ps.h
kynex7510: The name comes from 3ds_ps; 0x10001 happens to be the only public exponent used officially, and it is passed as a u32, but IMO it would be wrong to enforce a public-small/private-big grouping for exponents, as there's no actual constraint for what concerns the RSA engine, and it's purely a conventional distinction.

TuxSH: 0x10001 is the most used RSA public exponent _ever_. No one is going to use small exponents for private exponent as it would be trivially factored.

Does the field has any impact on p9?

And "Tiny ps improvement" isn't an acceptable commit name, Nikki.

kynex7510: Realistically speaking no one is going to use this API anyway, as it's limited in use case with very specific purposes; my argument for keeping things as generic as possible, and the reason why I made this PR, is documentation; as a reverse engineer myself I think libctru is a valuable resource even outside programming, and enforcing such convention on the structure would pass off the (incorrect) message that the engine makes distinctions between exponent types, when the distinction is merely about encoding.

TuxSH: > Realistically speaking no one is going to use this API anyway

This is not an excuse to commit technical debt

The problem with `is_full_exponent` is that it is much less clear than `is_private_key` for a user wanting to use this API. It also encourages misuse of the API, as 4-byte private keys would be very trivially bruteforced.

> as a reverse engineer myself I think libctru is a valuable resource even outside programming, and enforcing such convention on the structure would pass off the (incorrect) message that the engine makes distinctions between exponent types

So am I. You could rename the field as advised then explain what it means in the doxygen comment (like you already did).

Issue #590, opened 2025-09-14
Regression in console row scrolling (v2.6.2): first row sticky


MoeMorox:
### What's the issue you encountered?

Console scrolling no longer works correctly in `libctru 2.6.2`.  
- The very first printed line becomes sticky in the first row of the console.  
- Starting from the second line, scrolling happens normally, but the first line never scrolls off screen as it should.  

No changes were made to devkitPro tooling besides updating to `libctru 2.6.2-1` from pacman.  

### How can the issue be reproduced?

1. Initialize the console:  
   ```c
   gfxInitDefault();
   consoleInit(GFX_TOP, NULL);
   for (int i = 0; i < 100; i++) {
       printf("Line %d\n", i);
   }
   ```
2. Run the program until more lines than the screen can fit are printed.

3. Observe that the first line remains stuck at the top row.

### Environment?

Host OS: WSL Ubuntu (6.6.87.2-microsoft-standard-WSL2)

libctru version: 2.6.2-1 (installed via pacman)

devkitARM: latest from pacman (unchanged)

Tested on:

Nintendo 3DS hardware

Azahar emulator

### Additional context?

Screenshots / GIFs showing the issue:

bug_scroll_1.gif

bug_scroll_2.gif

Minimal repro program:
```c
#include <3ds.h>
#include <stdio.h>

int main(int argc, char **argv)
{
    gfxInitDefault();
    consoleInit(GFX_TOP, NULL);

    int counter = 0;

    while (aptMainLoop())
    {
        hidScanInput();
        u32 kDown = hidKeysDown();

        if (kDown & KEY_START)
            break;

        printf("Zeile %d\n", counter++);

        gspWaitForVBlank();
        gfxSwapBuffers();
    }

    gfxExit();
    return 0;
}
```
v2.6.1:
![Image](https://github.com/user-attachments/assets/1700f35c-ab89-4b4b-94c8-78e76f0b393f)
v.2.6.2:
![Image](https://github.com/user-attachments/assets/227c70b7-916c-41ef-9fdc-ddfa5ed23ea2)

PR #596, opened 2025-12-31
Fix console scrolling with sticky first row

Gustavo-Blois:
First row in console was sticky due to a bug reported in issue #590


Gustavo-Blois:
https://github.com/user-attachments/assets/1dba5c6c-3311-47e2-ac46-26f905c16001

